PR #3, opened 2020-07-03
Write some initial parts of the tutorial

ghost:


src/getting-started/hello-world.md
```diff
@@ -1,3 +1,186 @@
...
+/// The HelloWorld "class"
+#[derive(NativeClass)]
+#[inherit(Node)]
+pub struct HelloWorld;
+
+// Function that registers all exposed classes to Godot
+fn init(handle: InitHandle) {
+    // Register the new `HelloWorld` type we just declared.
+    handle.add_class::<HelloWorld>();
+}
```
karroffel, line 112: This code does not compile yet for multiple reasons. That's fine, but maybe note down that chunks of code are not meant to compile.

I remember in the Amethyst tutorial it's very confusing because the tutorial makes you write code that errors at runtime and then *afterwards* tell you to correct it. I spent a bit of time trying to debug it on my own without continuing in the tutorial, so maybe a note might be good here?

Maybe something like
> We will go step by step and discover what's needed to create script "class". Intermediate code versions might not compile but at the end of this section the code should compile.

src/getting-started/hello-world.md
```diff
@@ -1,3 +1,186 @@
...
+│   ├   Cargo.toml
+├─── my-godot-project
+│   ├─── .import
+│   └─── assets
+│   │   ├   icon.png
+│   │   └   icon.png.import
+│   └   project.godot
+```
+
+Once the project is creates, open `Cargo.toml`, change the crate type to `cdylib`, and add `gdnative` as a dependency:
```
halzy: `creates` => `created`

src/getting-started/hello-world.md
```diff
@@ -1,3 +1,186 @@
...
+use gdnative::prelude::*;
+
+// Function that registers all exposed classes to Godot
+fn init(handle: InitHandle) {
+}
+
+// Macros that create the entry-points of the dynamic library.
+godot_gdnative_init!();
+godot_nativescript_init!(init);
+godot_gdnative_terminate!();
```
halzy, line 64: What do you think of reducing these 3 lines to `godot_init!(init)`?

Most users of this crate will never pass in callbacks to the other 2 functions. Does it help them to know they exist? Perhaps it's for an advanced section of the book?

src/getting-started/hello-world.md
```diff
@@ -1,3 +1,186 @@
...
+
+Here, the `#[export]` attribute is used to tell godot-rust to expose your methods to Godot. In this case, we are overriding [`_ready`](https://docs.godotengine.org/en/stable/classes/class_node.html#class-node-method-ready) and printing a line of text.
+
+Now, re-compile the crate using `cargo build` and copy the resulting binary to the Godot project. Launch the project from the editor, and you should see `Hello, world!` in the Godot console!
+
+## Wrapping it up
+
+Congratulations! You have just created your first Rust GDNative library. You have learned how to expose scripts, methods to Godot using the bindings, and how to use them in Godot. A lot of the details are still unexplained, but you're off to a good start!
+
+You can find the full code for this examplein the main repo: https://github.com/godot-rust/godot-rust/tree/master/examples/hello_world
```
halzy: `examplin` => `example in`

halzy (commented):
It's looking good! Just a couple small typos.

src/getting-started/hello-world.md
ghost: Maybe it's a good idea to introduce a convenience `godot_init` macro? @karroffel What do you think?

ghost:
Updated according to comments!

src/getting-started/hello-world.md
karroffel: I guess that's fine, it is really rare that somebody does customisation of anything else but the `nativescript_init` hook so far.

ghost: Made a PR for this: https://github.com/godot-rust/godot-rust/pull/522

ghost:
522 is merged now, so I guess this is good to go?

halzy (approved):

PR #9, opened 2020-08-21
Added some first FAQ entries

bluenote10:
To make it more easy to add content to the user guide I thought I'll start a kind of FAQ section. Because of the assorted nature of an FAQ, knowledge can be added without worrying too much about arranging it in a meaningful order. Maybe some parts can be restructured later.

So far I have only added topics I had been involved with, mainly because they are the only ones I can roughly answer ;). My answers are still largely guesses, and may need some refinement.

I guess many of the [issues marked as question](https://github.com/godot-rust/godot-rust/issues?q=label%3Aquestion+) are good candidates for the FAQ as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,17 +2,19 @@
 
 The godot-rust book is a user guide for the Rust bindings. The book is currently an early work-in-progress, and contributions are very welcome.
 
-The book is built with [mdBook](https://github.com/rust-lang-nursery/mdBook). There is currently no online version available. To build the book locally, you can run:
+An online version of the book is available at [godot-rust.github.io/book/](https://godot-rust.github.io/book/).
+
+The book is built with [mdBook](https://github.com/rust-lang-nursery/mdBook). To build the book locally, you can run:
 
 ```
 $ cargo install mdbook
-$ mdbook build book
+$ mdbook build
 ```
 
 To run a local server with automatic updates while editing the book, run:
 
 ```
-$ mdbook serve book
+$ mdbook serve
 ```
 
 This repository is for documentation only. Please make pull requests for the library itself in the [main repo](https://github.com/godot-rust/godot-rust).
--- a/src/faq.md
+++ b/src/faq.md
@@ -1,0 +1,114 @@
+# FAQ
+
+## Passing additional arguments to a class constructor
+
+**Question**
+
+A native script type needs to implement `fn new(_owner: &Node) -> Self`.
+Is it possible to pass additional arguments to `new`?
+
+**Answer**
+
+Unfortunately this is currently a general limitation of GDNative (see [related issue](https://github.com/godotengine/godot/issues/23260)).
+
+As a result, a common pattern to work-around the limitation is to use explicit initialization methods. For instance:
+
+```rust
+#[derive(NativeClass)]
+#[inherit(Object)]
+struct DataWrapper {
+    data: Option<Data>,
+}
+
+#[godot::methods]
+impl DataWrapper {
+    fn new(_owner: &Object) -> Self {
+        DataWrapper {
+            data: None,
+        }
+    }
+
+    #[export]
+    fn set_data(&mut self, _owner: &Object, additional_arg1: i32, additional_arg2: i32) {
+        self.data = Some(Data::new(additional_arg1, additional_arg2));
+    }
+}
+```
+
+
+## Static methods
+
+**Question**
+
+In GDScript, classes can have static methods.
+However, when I try to omit `self` in the exported method signature, I'm getting a compile error.
+How can I implement a static method?
+
+**Answer**
+
+This is another limitation of GDNative -- static methods are not supported in general.
+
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -2,17 +2,19 @@
...
 
-The book is built with [mdBook](https://github.com/rust-lang-nursery/mdBook). There is currently no online version available. To build the book locally, you can run:
+An online version of the book is available at [godot-rust.github.io/book/](https://godot-rust.github.io/book/).
+
+The book is built with [mdBook](https://github.com/rust-lang-nursery/mdBook). To build the book locally, you can run:
 
 ```
 $ cargo install mdbook
-$ mdbook build book
+$ mdbook build
```
bluenote10, line 11: Note that if a user `cd`s to the cloned folder (which is the natural thing to do) the existing command doesn't work. At first glance I thought `book` refers to a command of `mdbook`, and took a second to figure out that is meant to be a path. Perhaps it is less confusing to specify the path.

ghost: This was leftover from when the guide was a subdirectory of the main repo. Thanks!

src/faq.md
```diff
@@ -0,0 +1,114 @@
+# FAQ
+
+## Passing additional arguments to a class constructor
+
+**Question**
+
+A native script type needs to implement `fn new(_owner: &Node) -> Self`.
```
ghost: I think it's best to remove the `_` prefix here, since it's talking about the signature in general without a specific function body? The underscore is for silencing unused variable warnings when the owner isn't used in the body.

src/faq.md
```diff
@@ -0,0 +1,114 @@
...
+**Answer**
+
+This is another limitation of GDNative -- static methods are not supported in general.
+
+As a work-around, it is possible to use a ZST (zero-sized type):
+
+```rust
+#[derive(NativeClass)]
+#[inherit(Object)]
+pub struct StaticUtil;
```
ghost, line 57: Might be worth it to add `#[derive(Copy, Clone, Default)]` and `#[user_data(Aether<StaticUtil>)]` here, and note that it's a specialized userdata wrapper for ZSTs? https://docs.rs/gdnative/0.9.0-preview.0/gdnative/prelude/struct.Aether.html

src/faq.md
```diff
@@ -0,0 +1,114 @@
...
+- as a [singleton auto-load object](https://docs.godotengine.org/en/stable/getting_started/step_by_step/singletons_autoload.html).
+
+
+## Converting a Godot type to the underlying Rust type
+
+**Question**
+
+I have a method that takes an argument `my_object` as a `Variant`.
+I know that this object is actually another native script, i.e., it is backed by some Rust data structure, called say `MyObject`.
+How can I convert the `Variant` to the underlying Rust type?
```
ghost: Might be better to use precise terminology here? e.g.:

> I know that this object has a Rust native script attached to it, called say `MyObject`. How can I access the Rust type given the `Variant`?

src/faq.md
```diff
@@ -0,0 +1,114 @@
...
+
+**Answer**
+
+This conversion can be accomplished by casting the `Variant` to a `Ref`, obtaining the underlying `RefInstance`, and mapping over it to access the Rust data type:
```
ghost: Converting from `Ref` / `TRef` to `Instance` / `RefInstance` is also considered casting:

> This conversion can be accomplished by casting the `Variant` to a `Ref`, and then to an `Instance` or `RefInstance`, and mapping over it to access the Rust data type:

ghost (commented):
Thanks for the PR! I believe this is a valuable addition to the guide. There are a few minor places that I think may be improved:

src/faq.md
```diff
@@ -0,0 +1,114 @@
...
+                .expect("Failed to convert my_object variant to object")
+                .assume_safe()
+        };
+        // 2. Obtain a RefInstance.
+        let my_object = my_object
+            .cast_instance::<MyObject>()
+            .expect("Failed to cast my_object object to instance");
+        // 3. Map over the RefInstance to extract the underlying user data.
+        my_object
+            .map(|my_object, _| {
```
ghost: Maybe the second argument can be named `_owner` so that those reading this page can understand what it is?

bluenote10:
Thanks, all very good improvements! I've pushed an update.

src/faq.md
```diff
@@ -0,0 +1,118 @@
...
+
+**Answer**
+
+This is another limitation of GDNative -- static methods are not supported in general.
+
+As a work-around, it is possible to use a ZST (zero-sized type):
+
+```rust
+#[derive(NativeClass)]
+#[derive(Copy, Clone, Default)]
```
ghost: These two lines should probably be merged. Other than that I think this is good to go!

bluenote10: Yes, I was wondering about that as well, just thought you prefer it like that from the way you had written. Fixed now ;)

ghost (approved):

ghost:
Merged. Thanks!

PR #16, opened 2020-09-25
Misc fixes in Getting Started

akien-mga:
- Make links clickable
- Fix default structure of Godot project
- Fix plural in reference to macros from earlier versions
- Typo fixes

---

Some of the changes made might be up for discussion / a matter of taste, so feel free to tell me what you prefer to keep unchanged, and I'll amend the commit.

### DIFF
--- a/src/getting-started/hello-world.md
+++ b/src/getting-started/hello-world.md
@@ -2,7 +2,7 @@
 
 Follow this tutorial to learn how to create an empty project that simply prints "Hello, world!" to the Godot console on ready. The code might not compile or work as intended while it's in-progress, but at the end of this section, the code will be compiling and working fine.
 
-The full, finished code is available in the main repo: https://github.com/godot-rust/godot-rust/tree/master/examples/hello_world
+The full, finished code is available in the main repo: [https://github.com/godot-rust/godot-rust/tree/master/examples/hello_world](https://github.com/godot-rust/godot-rust/tree/master/examples/hello_world).
 
 ## Creating the project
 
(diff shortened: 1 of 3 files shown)


akien-mga:
Another thing which I did not change here but noticed while reading: in https://github.com/godot-rust/book/blob/master/src/faq.md the `#[godot::methods]` attribute is used, while the (likely more up-to-date) Hello World tutorial uses only `#[methods]`. Is the former an earlier variant that should now be changed?

karroffel:
Thanks a lot for the PR!

> the `#[godot::methods]` attribute is used, while the (likely more up-to-date) Hello World tutorial uses only `#[methods]`.

That's just a matter of style, all of these are equivalent in terms of Rust attributes
```rust
#[gdnative::methods]
impl MyType {...}
```
```rust
use gdnative::*;

#[methods]
impl MyType {...}
```

```rust
extern crate gdnative as godot;

#[godot::methods]
impl MyType {...}
```

That said, it should probably be made consistent, thanks for pointing that out!

src/getting-started/hello-world.md
```diff
@@ -83,15 +82,15 @@ This declares an empty callback function, which is called when the library is lo
 godot_init!(init);
 ```
 
-These macros define the necessary C callbacks used by Godot. You only need *one* of these in the entire library. Note how the `init` function defined earlier is given to the `godot_nativescript_init` macro as a callback.
+This macro defines the necessary C callbacks used by Godot. You only need *one* of call to this macro in the entire library. Note how the `init` function defined earlier is given to the `godot_init!` macro as a callback.
```
karroffel:
```suggestion
This macro defines the necessary C callbacks used by Godot. You only need *one* invocation of this macro in the entire library. Note how the `init` function defined earlier is given to the `godot_init!` macro as a callback.
```

karroffel (changes requested):
Nice catches! I made one small suggestion, other than that it should be good to go!

src/getting-started/hello-world.md
akien-mga: Thanks, fixed.

karroffel (approved):
Looks fine to me, thanks a lot!

PR #21, opened 2021-02-18
Be more clear with the maintenance state of the template

ghost:


ghost:
This marks the template more clearly as a WiP since Windows support is still not quite in place, and we have users encounter problems with the setup.

PR #22, opened 2021-03-07
Page about data exchange GDScript <-> Rust

Bromeon:
It took me quite some time to understand all the different mechanisms for data exchange between GDScript and Rust. With this documentation, I'd like to share the gained knowledge and contribute to a more coherent understanding.

The writing is not entirely complete yet, I planned to add an example about `Object.call()`, but I thought it would be good to already submit a first draft and get some feedback. Also, it's very well possible that I misunderstood some parts; don't hesitate to correct me in that case.

One thing where I would like to hear the opinion of the developers, is when to use `Ref` and when `Instance`. Is it simply "when the class is a user-defined script, use instance"?

**[Edit 2]:** split into two pages; rendered versions:
* [Data representations](https://github.com/godot-rust/book/blob/27886573bfbdb0dbf5e8fc929b423fdb46c91d23/src/advanced-guides/data-representations.md)
* [Exchanging data between GDScript and Rust](https://github.com/godot-rust/book/blob/27886573bfbdb0dbf5e8fc929b423fdb46c91d23/src/advanced-guides/data-exchange.md)


src/advanced-guides/data-exchange.md
```diff
@@ -0,0 +1,386 @@
...
+* [Passing data from Rust to GDScript](#passing-data-from-rust-to-gdscript)  
+  * [Class construction in Rust](#class-construction-in-rust)
+  
+_For readability, code blocks use the `gd` module name in place of `gdnative`. Its submodules are listed with full name, for clarity._  
+_In your own code, you may consider to use the prelude:_
+```rust
+mod gd {
+    pub use gdnative::prelude::*;
+}
+```
```
ghost: Throughout the user guide we simply suggest that users glob import `prelude`, so I'd suggest removing this part and the `gd::` prefix in any code examples.

src/advanced-guides/data-exchange.md
```diff
@@ -0,0 +1,386 @@
...
+gd::godot_nativescript_init!(init);
+gd::godot_gdnative_terminate!();
+```
+
+The argument `init` refers to the function registering native script classes, which is also defined by you. Let's assume you have one class `GodotApi`, which exposes a public interface to be invoked from Godot. The registration is then as follows:
+```rust
+fn init(handle: gd::nativescript::InitHandle) {
+	handle.add_class::<GodotApi>();
+}
+```
```
ghost: This information really belongs in the tutorial, instead of an "advanced guide", imo. I'd suggest checking what exists in the tutorial and incorporating any missing parts there.

src/advanced-guides/data-exchange.md
```diff
@@ -0,0 +1,386 @@
...
+#[gd::methods]
+impl GodotApi {
+    // Constructor, either:
+    fn new(owner: &gd::api::Node) -> Self { ... }
+    // or:
+    fn new(owner: gd::TRef<gd::api::Node>) -> Self { ... }
+}
+```
+
+The function `new()` corresponds to `_init()` in GDScript. The _owner_ is the base object of the script, and must correspond to the class specified in the `#[inherit]` attribute (or `gd::api::Reference` if the attribute is absent). The parameter can be a shared reference `&T` or a `TRef<T>`.
```
ghost: The annotated code is great, perhaps move them into the hello world tutorial itself?

src/advanced-guides/data-exchange.md
```diff
@@ -0,0 +1,386 @@
...
+}
+```
+
+The function `new()` corresponds to `_init()` in GDScript. The _owner_ is the base object of the script, and must correspond to the class specified in the `#[inherit]` attribute (or `gd::api::Reference` if the attribute is absent). The parameter can be a shared reference `&T` or a `TRef<T>`. 
+
+With a `new()` method, you are able to write `GodotApi.new()` in GDScript. If you don't need this, you can add the `#[no_constructor]` attribute to the struct declaration.
+
+At this point, arguments cannot be passed into the constructor. There are two main workarounds:
+* Declare an `initialize()` method or similar -- this however requires having uninitialized state for a short time, which may decrease type safety (e.g. if you need to use `Option` for a field, before assigning a value)
+* Register a separate class, which returns fully-constructed instances. So instead of `Enemy.new()` you can write `EnemyFactory.create(args)`.
```
ghost: > this however requires having uninitialized state for a short time, which may decrease type safety (e.g. if you need to use `Option` for a field, before assigning a value)

This is not true. You never have uninitialized state in safe code, and `Option` does not decrease type safety. What it *does* decrease is ease of use.

> Register a separate class, which returns fully-constructed instances. So instead of `Enemy.new()` you can write `EnemyFactory.create(args)`.

Perhaps a FAQ entry instead? I think users that need this will be less likely to find this information if it's buried in an "advanced guide".

src/advanced-guides/data-exchange.md
```diff
@@ -0,0 +1,386 @@
...
+
+
+## Exported methods
+
+In order to pass data from Godot to Rust, you can export methods. With the `#[export]` attribute, godot-rust takes care of method registration and serialization. Note that the constructor is not annotated with `#[export]`.
```
ghost: > In order to pass data from Godot to Rust, you can export methods. With the `#[export]` attribute

"In order to receive data from Godot", perhaps?

src/advanced-guides/data-exchange.md
```diff
@@ -0,0 +1,386 @@
...
+        self.enemy_count += 1;
+    }
+
+    #[gd::export]
+    fn count_enemies(&self, _owner: &gd::api::Node) -> i32 {
+        self.enemy_count
+    }  
+}
+```
+These two methods are semantically equivalent, yet they demonstrate how godot-rust implicitly converts the values to the parameter types (unboxing). You could use `Variant` everywhere, however it is more type-safe and expressive to use specific types. The same applies to return types, you could use `Variant` instead of `i32`.
```
ghost: It's not really "unboxing", but more akin to "marshalling" or even "serialization" at times, since `FromVariant` can do arbitrarily complex conversions.

src/advanced-guides/data-exchange.md
```diff
@@ -0,0 +1,386 @@
...
+    #[gd::export]
+    fn count_enemies(&self, _owner: &gd::api::Node) -> i32 {
+        self.enemy_count
+    }  
+}
+```
+These two methods are semantically equivalent, yet they demonstrate how godot-rust implicitly converts the values to the parameter types (unboxing). You could use `Variant` everywhere, however it is more type-safe and expressive to use specific types. The same applies to return types, you could use `Variant` instead of `i32`. 
+`GodotString` is the Godot engine string type, but it can be converted to standard `String`. To choose between the two, consult [the docs](https://docs.rs/gdnative/latest/gdnative/core_types/struct.GodotString.html).  
+
+In GDNative, you can then write this code:
```
ghost: Typo: In GDScript

src/advanced-guides/data-exchange.md
```diff
@@ -0,0 +1,386 @@
...
+
+#[gd::export]
+fn _process(&mut self, _owner: &gd::Node, delta: f32) {...}
+
+#[gd::export]
+fn _physics_process(&mut self, _owner: &gd::Node, delta: f32) {...}
+
+#[gd::export] // extremely useful
+fn _to_string(&self, _owner: &gd::Reference) -> String {...}
+```
```
ghost: Up till now what I've read really felt like a series of FAQ entries instead of a focused, "advanced" topic guide. I think it probably has to do with the book's structure not being very clearly defined, making it hard for you to find where such things need to live?

I'll probably have to spend more thoughts on organizing this later.

src/advanced-guides/data-exchange.md
```diff
@@ -0,0 +1,386 @@
...
+    enemy_count: i32,
+}
+```
+
+The GDScript code would be changed as follows.
+```python
+print("enemies: ", api.enemy_count)
+```
+
+That's it.
```
ghost: Isn't this less about "Passing data from GDScript to Rust", but more about "Exposing data from Rust to GDScript"?

src/advanced-guides/data-exchange.md
```diff
@@ -0,0 +1,386 @@
...
+
+
+## `Ref` and `TRef` smart pointers
```
ghost: I'm feeling like that this section deserves its own page / chapter, since it's well written, but really about using the smart pointer types *from Rust*: despite being under "passing data from GDScript to Rust" there isn't even any GDScript here.

src/advanced-guides/data-exchange.md
```diff
@@ -0,0 +1,386 @@
...
+    // ...
+    handle.add_class::<Enemy>();
+}
+```
+Now, it's not possible to directly return `Enemy` instances in exported methods, so this won't work:
+```rust
+#[gdnative::export]
+fn create_enemy(&mut self, _owner: &gd::Node) -> Enemy {...}
+```
+One way would be to implement the `ToVariant` trait, but much easier is to wrap the object in a `Instance<Enemy, Unique>`.
```
ghost: It should be noted that the two approaches lead to drastically different semantics:

- `ToVariant` maps to "plain old data" objects in GDScript. They *do not* have consistent identities. Two objects that are structurally identical but live at different addresses in GDScript will be considered equal when they are converted back to Rust values. Conversely, `Variant`s seen as equal in GDScript may convert into non-equal Rust values due to mutation.
- `Instance`s are objects with script instances. They have consistent identities, no matter where and how you modify them.
- The cost of `ToVariant` scales with the size and depth of the structure, since you're converting everything to their Godot counterparts. However, the upside is that it results in native structures that can be easily used and manipulated in GDScript.
- Conversely, moving a Rust value into an `Instance` has a much flatter cost, but the resulting values can't be used as easily. You'll have to expose methods for accessing and/or modifying the data. This can, however, be beneficial if you want to limit how they can be used from GDScript, or if the data is hard to consume (for GDScript) in raw form.

So the situation is not as simply cut as "`Instance` is easier" here. It should really depend on what kind of semantics you want, and maybe performance characteristics, if deemed relevant after profiling.

src/advanced-guides/data-exchange.md
```diff
@@ -0,0 +1,386 @@
...
+) -> gd::Instance<Enemy, gd::thread_access::Unique> {
+    let enemy = Enemy {
+        pos: gd::Vector2::new(7.0, 2.0),
+        health: 100.0,
+        name: "MyEnemy".to_string(),
+    };
+    
+    gd::NativeClass::emplace(enemy)
+    // or, with 'use gd::nativescript::NativeClass;'
+    Enemy::emplace(enemy)
```
ghost: Or `enemy.emplace()` if you imported the trait (in the prelude)?

ghost (commented):
Thanks for the PR! The comments below mostly represents my thoughts as I read through the article, and are just suggestions from the perspective of a reader. I think most of the information here is very desirable, and there is a visible effort put into the docs, but I felt that the "Exchanging data between GDScript and Rust" page is perhaps a bit unfocused in the topics it deals with.

I consider the book a collaborative document: instead of a collection of pages each owned by some individual, it should be viewed as a whole thing owned by the community. As such, contributors are not limited to the one or few pages they "own". Instead, when you find any part of the guide inadequate, you can go ahead and just modify those pages. In the comments for "Exchanging data between GDScript and Rust" I made some suggestions as of where bits and pieces of knowledge might better belong.

"Data representations", on the other hand, feels very coherent. I like it a lot!

All comments are just suggestions here.

src/advanced-guides/data-exchange.md
Bromeon: Ok, I can change that. The idea was to make clear which classes belong to godot-rust, but I think there are no ambiguities. What about the submodules like `api` etc? They would be removed with the prelude as well, but I'm not sure if this is then not annoying for users who want their identifiers qualified.

I see it a bit like this:
* if someone wants to not use module prefixes, he can simply omit them
* if someone is strict about using them, he needs to look up every symbol in the doc and check their module (as all the code examples won't have them).

I'm probably overthinking it though, and shorter code is better. Let me know what is preferred.

src/advanced-guides/data-exchange.md
Bromeon: see main comment

src/advanced-guides/data-exchange.md
Bromeon: > This is not true. You never have uninitialized state in safe code, and Option does not decrease type safety. What it does decrease is ease of use.

Bad wording on my side. How do you call things that are technically initialized (as in safe), but not semantically initialized (an empty `Option` shell instead of the value)? Basically, this is the Java problem where everything may be null and you need to constantly `unwrap()` or `if let` before accessing fields that should always have a value.

> Perhaps a FAQ entry instead? 

Good point. Then I could link to that FAQ entry from here, to make sure people only reading this will not miss it.

src/advanced-guides/data-exchange.md
Bromeon: 👍🏼

src/advanced-guides/data-exchange.md
Bromeon: 👍🏼

src/advanced-guides/data-exchange.md
Bromeon: 👍🏼

src/advanced-guides/data-exchange.md
Bromeon: see main comment

src/advanced-guides/data-exchange.md
Bromeon: I was thinking from a data flow perspective: you have some state in Godot that you want to bring into Rust. But I agree that this property is exactly the opposite direction -- would be different for a `set` property.

Maybe the "data flow" point of view is not the best way to categorize things.

Bromeon:
(Main comment)

> I consider the book a collaborative document: instead of a collection of pages each owned by some individual, it should be viewed as a whole thing owned by the community. As such, contributors are not limited to the one or few pages they "own". Instead, when you find any part of the guide inadequate, you can go ahead and just modify those pages.

Absolutely. I was also quite unsure if _Advanced Guides_ is the right place, or about organization in general. The idea was to provide some basic content, and then get feedback on improvements and re-organization. Of course I appreciate if people improve on this in the future, I don't consider it mine or I would have posted it on a personal blog 🙂 

---

Regarding duplication of some parts with the Tutorial/Hello World:
Yes, that also went through my head. But a large part of the page should be considered "basic" as in "you need to know these things to reasonably work with godot-rust". So I'm not even sure if _Advanced Guides_ is the right section for it, or where you draw the line.

On the other hand, I found myself struggling with the very problem that a lot of information is scattered across documentation, FAQ, the book and examples. This page intends to be an in-depth explanation of all the possible data exchange mechanisms (not yet complete) and answer questions like "I want to call a custom function in GDScript, what is the background I need to know and how can I make it work".

I don't know what the ideal structure for this would be, now it's a bit a mix of advanced tutorial and reference (ad-hoc lookup). But I fear if we move a lot of the "basic stuff" like registering classes or exporting methods to the first tutorial, it a) feels like some parts are missing in the "complete reference" and b) we cannot go into greater detail of behind-the-scenes mechanics, as that would defeat the purpose of a simple introduction to get something on screen.

But it also doesn't have to be one single page, it should just be immediately clear where to find all the information.

src/advanced-guides/data-exchange.md
Bromeon: Isn't [`Instance::emplace()`](https://docs.rs/gdnative/0.9.3/gdnative/nativescript/struct.Instance.html#method.emplace) an associated method, or is there something else declared implicitly?

src/advanced-guides/data-exchange.md
Bromeon: Since you seem to know this aspect much better than me, would it make sense if you explained trade-offs directly in the document? Or should I copy some of your text?

We should probably also explain the "plain old data" concept and identities, maybe in the other page?

src/advanced-guides/data-exchange.md
ghost: `Instance::emplace` is associated, but what you're invoking here is [`NativeClass::emplace`](https://docs.rs/gdnative/0.9.3/gdnative/nativescript/trait.NativeClass.html#method.emplace), a trait method with the same name, which takes `self` and thus can be called this way.

src/advanced-guides/data-exchange.md
Bromeon: True, but you don't get around `Ref/TRef` when you want to pass classes from GDScript to Rust and the chapter explains them in that context (not as a general-purpose documentation).

But this might also become more clear once we have a better idea on the overall structure and organization (main comment).

src/advanced-guides/data-exchange.md
ghost: `prelude`s are modules that are explicitly intended to be glob imported by users for convenience. Since we do provide a `prelude`, it's important that we be consistent in our attitude towards it. `api` types should just be imported individually in examples if necessary, although they should be replaced by those already re-exported in `prelude` unless really necessary.

> I'm not sure if this is then not annoying for users who want their identifiers qualified.

The other side should also be considered: removing module prefixes is annoying for those who do import `prelude`.

> if someone is strict about using them, he needs to look up every symbol in the doc and check their module (as all the code examples won't have them).

The prelude exists because there are too many types and traits necessary to use the library to import individually, and importing it is the recommended way to use the library. I'd consider it fine to make it harder to use the library in the non-recommended way.

src/advanced-guides/data-exchange.md
ghost: > How do you call things that are technically initialized (as in safe), but not semantically initialized (an empty Option shell instead of the value)?

I'm not really sure either. Perhaps just reword it somehow so we don't have to give a name to the situation? e.g. `` Declare an `initialize()` method or similar -- this however may require keeping `Option`s in the struct if you cannot assign sensible defaults, and can make the code more tedious to write ``.

src/advanced-guides/data-exchange.md
ghost: > would it make sense if you explained trade-offs directly in the document? Or should I copy some of your text?

Either way you like. You can copy the text or reword it if you want. I can also push changes directly if you enabled the option that allowed maintainers to push to your branch.

> We should probably also explain the "plain old data" concept and identities, maybe in the other page?

That makes sense!

src/advanced-guides/data-exchange.md
Bromeon: Ok, so I'll remove **all** module qualifiers. Let me know if you would like to keep some like `api`.

src/advanced-guides/data-exchange.md
Bromeon: 👍🏼

src/advanced-guides/data-exchange.md
Bromeon: Oh sorry for that, it's enabled now!

Do you know a place to look for this POD and identity topic, maybe Godot's own documentation? Or do you already know the things off the top of your head?

src/advanced-guides/data-exchange.md
Bromeon: Ah, I see. Maybe I'll explicitly explain how `NativeClass` methods are generated, so it doesn't appear like magic to the user.

ghost:
> Yes, that also went through my head. But a large part of the page should be considered "basic" as in "you need to know these things to reasonably work with godot-rust". So I'm not even sure if Advanced Guides is the right section for it, or where you draw the line.

Perhaps _An Overview of GDNative_ could be used for this purpose? _Data representations_ certainly looks like it could reside there. I'm also not sure where we should draw the line for _Advanced Guides_, but we probably can agree that it should be for things that are intended for niche audiences. Then I guess we can draw the line at "if it's useful for everyone, it does not belong there", I suppose?

> On the other hand, I found myself struggling with the very problem that a lot of information is scattered across documentation, FAQ, the book and examples. ...But it also doesn't have to be one single page, it should just be immediately clear where to find all the information.

Sure, that is what the user guide is created and should be used for. Perhaps "data exchange" is just too broad a topic to cover in a single page, and that's what made it feel unfocused to me. Regarding the tutorial, I think we'll want to expand on it further and introduce more advanced topic gradually in the future, but personally I don't have enough time for that right now, so sadly it will have to stay in the limbo for a bit unless someone else comes along.

For collecting the "reference" in one place, I suppose we can probably separate the article into multiple pages, each about the usage of one thing -- methods, properties, signals, and so on, and put them in a new top-level category, possibly named _Exposing a GDScript API_? I think most of the page right now can be summarized under that name, but please feel free to suggest one if you have a better idea for the category.

src/advanced-guides/data-exchange.md
ghost: It was just from my memory, yes. I don't think it's discussed anywhere in Godot's docs -- referential equality is taken for granted in most languages, after all. I think it's worth noting since the difference in semantics can become a pitfall, for example, if you try to use things as dictionary keys.

Bromeon:
With "Exposing a GDScript API" you mean a Rust API to call _from_ GDScript or literally a "GDScript API"? 😉 

What if the mechanisms would be their top-level category in the book, i.e. not in _Advanced Guides_.

* Getting Started
  * Setup
  * Hello World
* **An Overview of GDNative** (not sure if this is how you meant it, but it could fit. alternative "Binding to Rust code")
  * Data representations
  * Ref, TRef and Instance -- wrappers for custom types
  * Class registration
  * Exported methods
  * Exported properties
  * (maybe the manual registration API of 0.10, signal export etc)
  * Calling into GDScript from Rust (includes call, signals, get/set)
* Advanced Guides
  * niche topics, without which you can still develop a good godot-rust game

With this, I'll need to explain _Ref/TRef/Instance_ before the concept of exported methods is taught, but in turn this allows each chapter about methods, properties etc. to be a fully complete reference with all possible use cases.

ghost:
I suppose I was thinking more like this?

* Getting Started
  * Setup
  * Hello World
* **An Overview of GDNative** (where GDNative means GDNative the technology, not `gdnative` the crate)
  * Data representations
  * Ref, TRef and Instance -- wrappers for custom types (here because it's about how Godot handles objects and scripts)
  * (perhaps some more info as necessary)
* Binding to Rust code (as you suggested)
  * Class registration
  * Exported methods
  * Exported properties
  * (maybe the manual registration API of 0.10, signal export etc)
  * Calling into GDScript from Rust (includes call, signals, get/set)
* Advanced Guides
  * niche topics, without which you can still develop a good godot-rust game

> in turn this allows each chapter about methods, properties etc. to be a fully complete reference with all possible use cases.

Yes and I think that would be pretty nice?

Bromeon:
That sounds good :)

I'll start the restructuring soon then, thanks a lot for the feedback! Apart from the reorganization, I will change these things:
* `NativeClass` generated methods -- https://github.com/godot-rust/book/pull/22#discussion_r590183072
* FAQ entry for constructor arguments -- https://github.com/godot-rust/book/pull/22#discussion_r589940892
* Remove all qualified modules -- https://github.com/godot-rust/book/pull/22#discussion_r589937981
* the couple of smaller typo/wording issues

And after that, I'll need your help here:
* POD and identity explanation -- https://github.com/godot-rust/book/pull/22#discussion_r589954623
* See if there's duplication with Hello World left and decide if we can live with it. Probably yes, maybe focus Hello World more on usage rather than technicalities/underlying concepts -- https://github.com/godot-rust/book/pull/22#discussion_r589939970

Bromeon:
Restructured everything, improved FAQ entry, added section about `call()` and `emit_signal()`, as well as standalone explanation of `Ref`/`TRef`/`Instance`.

I'm sure there are some errors left, but let me know what you think so far. Also, feel free to add commits to improve things directly.

ghost:
@Bromeon I think it's looking really good! :tada: I'll take a closer look and see if there are any mistakes later.

PR #24, opened 2021-03-20
Add more overview and introduction text

tom-leys:
Trying to help introduce users gradually to "Why" use godot-rust and more of a broad overview.

### DIFF
--- a/src/gdnative-overview.md
+++ b/src/gdnative-overview.md
@@ -1,12 +1,12 @@
 # An Overview of GDNative
 
-GDNative is the interface between the Godot engine and bindings in native languages, such as C, C++ or Rust. It defines how Godot finds and loads Dynamically Linked Libraries (DLLs / dylibs) and understands their contents. 
+GDNative is the interface between the Godot engine and bindings in native languages, such as C, C++ or Rust. It defines how Godot finds and loads your dylib (shared library / DLL) and understands its contents. 
 
 This chapter gives a broad overview of basic GDNative concepts and godot-rust's approach to implement them in Rust. It is not a usage guide for exposing your Rust code to Godot; see chapter [Binding to Rust code](rust-binding.md) for concrete examples.
 
 When working with GDNative you'll be primarily worried about:
-1. Node types such as Node or Spatial (using a type like TRef<Spatial>)  
-2. Resource types such as Mesh, Texture or PackedScene 
+1. `Node` types such as `Node` or `Spatial` (using a type like `TRef<Spatial>`)  
+2. `Resource` types such as `Mesh`, `Texture` or `PackedScene` 
 3. Data you can send to or receive from other scripts (perhaps in GDScript) such as Variant (eg: values, dictionaries or lists)  
 4. How to create your own types, supporting incoming calls.
 
--- a/src/gdnative-overview.md
+++ b/src/gdnative-overview.md
@@ -1,8 +1,14 @@
 # An Overview of GDNative
 
-GDNative is the interface between the Godot engine and bindings in native languages, such as C, C++ or Rust.
+GDNative is the interface between the Godot engine and bindings in native languages, such as C, C++ or Rust. It defines how Godot finds and loads Dynamically Linked Libraries (DLLs / dylibs) and understands their contents. 
 
 This chapter gives a broad overview of basic GDNative concepts and godot-rust's approach to implement them in Rust. It is not a usage guide for exposing your Rust code to Godot; see chapter [Binding to Rust code](rust-binding.md) for concrete examples.
+
+When working with GDNative you'll be primarily worried about:
+1. Node types such as Node or Spatial (using a type like TRef<Spatial>)  
+2. Resource types such as Mesh, Texture or PackedScene 
+3. Data you can send to or receive from other scripts (perhaps in GDScript) such as Variant (eg: values, dictionaries or lists)  
+4. How to create your own types, supporting incoming calls.
 
 Subchapters:
 
--- a/src/introduction.md
+++ b/src/introduction.md
@@ -10,7 +10,7 @@ If you have used earlier versions of `godot-rust` before, see [Migrating from go
 
 The Godot Engine has rich user scripting support. Using a DLL binding system called GDNative you can create "Scripts" in Rust and attach them to Godot Nodes. Everything you can do from Godot in a language such as GDScript you can do from rust using godot-rust.
 
-This DLL - short for Dynamically linked Library (rust calls it a dylib) also has full access to Rust's entire featureset. You can read and write files, open network ports, connect to a custom server, run a custom website and more.
+Your code is compiled into a dylib (often called a shared library or DLL) which is loaded by Godot and also has full access to Rust's entire featureset. You can read and write files, open network ports, connect to a custom server, run a custom website and more.
 
 Rust itself is a far more rigourous and complete language than GDScript, offering high level features such as traits or easy packing and unpacking of values. Rust also offers ridiculous performance, so if your game simulation is complex, rust is a good language to run it.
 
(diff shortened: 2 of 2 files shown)


src/gdnative-overview.md
```diff
@@ -1,9 +1,15 @@
 # An Overview of GDNative
 
-GDNative is the interface between the Godot engine and bindings in native languages, such as C, C++ or Rust.
+GDNative is the interface between the Godot engine and bindings in native languages, such as C, C++ or Rust. It defines how Godot finds and loads Dynamically Linked Libraries (DLLs / dylibs) and understands their contents.
```
ghost: "Dynamically Linked Libraries" seems too Windows-centric with the capitalization. Perhaps "dynamic libraries (DLLs / shared objects / dylibs)"?

src/gdnative-overview.md
```diff
@@ -1,9 +1,15 @@
 # An Overview of GDNative
 
-GDNative is the interface between the Godot engine and bindings in native languages, such as C, C++ or Rust.
+GDNative is the interface between the Godot engine and bindings in native languages, such as C, C++ or Rust. It defines how Godot finds and loads Dynamically Linked Libraries (DLLs / dylibs) and understands their contents. 
 
 This chapter gives a broad overview of basic GDNative concepts and godot-rust's approach to implement them in Rust. It is not a usage guide for exposing your Rust code to Godot; see chapter [Binding to Rust code](rust-binding.md) for concrete examples.
 
+When working with GDNative you'll be primarily worried about:
+1. Node types such as Node or Spatial (using a type like TRef<Spatial>)  
+2. Resource types such as Mesh, Texture or PackedScene
```
ghost: Nitpick: Type names should be surrounded by `` ` `` so they appear as code spans.

src/introduction.md
```diff
@@ -6,10 +6,24 @@ If you're new to `godot-rust`, try the [Getting Started](./getting-started.md) t
 
 If you have used earlier versions of `godot-rust` before, see [Migrating from godot-rust 0.8.x](./migrating-0-8.md) for a quick guided tour to the new API.
 
+## What is godot-rust
+
+The Godot Engine has rich user scripting support. Using a DLL binding system called GDNative you can create "Scripts" in Rust and attach them to Godot Nodes. Everything you can do from Godot in a language such as GDScript you can do from rust using godot-rust.
+
+This DLL - short for Dynamically linked Library (rust calls it a dylib) also has full access to Rust's entire featureset. You can read and write files, open network ports, connect to a custom server, run a custom website and more.
```
ghost: Same as previous comment about avoiding Windows-centricity.

src/introduction.md
```diff
@@ -6,10 +6,24 @@ If you're new to `godot-rust`, try the [Getting Started](./getting-started.md) t
...
+
+Rust itself is a far more rigourous and complete language than GDScript, offering high level features such as traits or easy packing and unpacking of values. Rust also offers ridiculous performance, so if your game simulation is complex, rust is a good language to run it.
```
ghost, line 15: Suggestion: "Rust also offers ridiculous performance" -- Being on par with C/C++ isn't necessarily ridiculous depending on what you're comparing it with, and it does sounds a bit like evangelism. Perhaps it makes sense to give the statement a little more context: "Despite its memory safety and high-level features, Rust offers very good performance that is on par with C/C++, by designing most of them as zero-cost abstractions"?

ghost (commented):
Thanks for the PR! Only a few nitpicks and suggestions about wording.

src/introduction.md
tom-leys: True. I kind of meant "using a compiled language like Rust, C/C++ makes performance a non-issue" - I also like your sentence. Possibly retain both.

src/introduction.md
tom-leys: Your code is compiled into a dylib (often called a shared library or DLL) which is loaded by Godot and also has full access to Rust's entire featureset.

src/gdnative-overview.md
tom-leys: "It defines how Godot finds and loads your dylib (shared library / DLL) and understands its contents."

src/introduction.md
Bromeon: Tiny nitpick, I always get shivers down my spine when I see "C/C++" 😬 
I would write "C and C++" to make clear they are distinct languages. I've seen enough people not knowing there's a difference...

Like toasteater, l would also write it a bit neutral, mentioning advantages of Rust without praising it too much 🙂

src/gdnative-overview.md
```diff
@@ -1,9 +1,15 @@
 # An Overview of GDNative
 
-GDNative is the interface between the Godot engine and bindings in native languages, such as C, C++ or Rust.
+GDNative is the interface between the Godot engine and bindings in native languages, such as C, C++ or Rust. It defines how Godot finds and loads your dylib (shared library / DLL) and understands its contents. 
 
 This chapter gives a broad overview of basic GDNative concepts and godot-rust's approach to implement them in Rust. It is not a usage guide for exposing your Rust code to Godot; see chapter [Binding to Rust code](rust-binding.md) for concrete examples.
 
+When working with GDNative you'll be primarily worried about:
+1. `Node` types such as `Node` or `Spatial` (using a type like `TRef<Spatial>`)
```
Bromeon, line 8: I would maybe not go into details about how to use them yet (the `TRef` example), because this page is just an overview and all the details will be explained in the subpages.

src/gdnative-overview.md
```diff
@@ -1,9 +1,15 @@
...
 
-GDNative is the interface between the Godot engine and bindings in native languages, such as C, C++ or Rust.
+GDNative is the interface between the Godot engine and bindings in native languages, such as C, C++ or Rust. It defines how Godot finds and loads your dylib (shared library / DLL) and understands its contents. 
 
 This chapter gives a broad overview of basic GDNative concepts and godot-rust's approach to implement them in Rust. It is not a usage guide for exposing your Rust code to Godot; see chapter [Binding to Rust code](rust-binding.md) for concrete examples.
 
+When working with GDNative you'll be primarily worried about:
+1. `Node` types such as `Node` or `Spatial` (using a type like `TRef<Spatial>`)  
+2. `Resource` types such as `Mesh`, `Texture` or `PackedScene` 
+3. Data you can send to or receive from other scripts (perhaps in GDScript) such as Variant (eg: values, dictionaries or lists)
```
Bromeon, line 10: maybe ``` `Variant` ``` (backticks)?

src/introduction.md
```diff
@@ -6,10 +6,24 @@ If you're new to `godot-rust`, try the [Getting Started](./getting-started.md) t
...
 
 ## License
 
-The `godot-rust` bindings, and this user guide, are licensed under the [MIT license](LICENSE.md).
\ No newline at end of file
+The `godot-rust` bindings, and this user guide, are licensed under the [MIT license](LICENSE.md).
+
+## Help
+
+There are several active discords which discuss Godot-Rust. [Here is an invite to the official one](https://discord.gg/AbjQp6mjKS).
```
Bromeon, line 29: This made me think, what is the official naming of the library?

I used lower-case "godot-rust" in the text I recently added, although it felt sometimes weird (it being a proper name), I think it's quite common for Rust libraries?

ghost: I agree that it's a bit complicated. While I don't suggest that this makes it official, personally I refer to it as the lower-case "godot-rust". As you have said, many Rust libraries do refer to themselves in lower-case. Maybe we can have a discussion about this later. For now, I suggest using "godot-rust" to be consistent with everything else.

jacobsky:
As this PR is a bit old and a lot has changed in the book, I will close this in favor of #43

PR #25, opened 2021-04-19
fix broken link in calling-gdscript

noam-stein:
the link was for:
`localhost:3000`

its should be for:
`https://godot-rust.github.io/book`


Bromeon:
I think it should be neither an Internet nor a localhost address, but rather a relative path. If you generate the documentation on your machine, the link should also follow there.

noam-stein:
you are right :-),
fixed

Bromeon:
Looks good to me.
Thank you! 🙂

PR #33, opened 2021-09-06
Updated the FAQ and changed some of the formatting.

### LINKED ISSUE #5: Frequently asked questions
- [ ] `Instance`
- [ ] User-data errors (`BorrowFailed`, `DifferentThread`)

...and more?

### LINKED ISSUE #23: Mention how to create gdnlib and gdns files

jacobsky:
@Bromeon As I mentioned via discord, I went through the issues and extrapolated on cleaned up the FAQ a bit.
Also added a fix to have the logging recipe get added to the summary as it was previously missing.


src/faq/code-faq.md
```diff
@@ -0,0 +1,312 @@
+# Common Code Questions
+
+## Avoiding a `BorrowFailed` error on method call
```
Bromeon: This should be removed.

src/faq/code-faq.md
```diff
@@ -0,0 +1,312 @@
...
+        // *self.data.lock().expect("this should work") = data;
+        owner.emit_signal("updated", &[]);
+    }
+```
+
+In both instances you will not encounter the reentrant errors.
+
+## Why are methods on all Godot API classes using `&self` even when they should cause mutations?
+
+1. Godot objects can be accessed and aliased freely but Rust's &mut expects only a single mutable alias at a time. In terms of type "safety" there is no specific type safety. We are aware that this may make it a little less obvious which
```
Bromeon: It's not very clear to me what you mean with this bullet point (the last sentence is incomplete). If it's about reduced type safety, I would probably mention it last.

To me the main message is along the following lines -- you already mention that in points 2 and 3, so those are good as-is:

> `&mut` means that only one reference can exist simultaneously (no aliasing), _not_ that the object is mutable. Mutability is often a _consequence_ of the exclusiveness. Since Godot objects are managed by the engine, Rust cannot guarantee that references are exclusive, as such using `&mut` would cause undefined behavior. Instead, an interior mutability pattern based on `&T` is used.

Maybe we can link again to https://docs.rs/dtolnay/latest/dtolnay/macro._02__reference_types.html (here with `latest` instead of a version)?

src/faq/code-faq.md
```diff
@@ -0,0 +1,312 @@
...
+
+2. In godot-rust, rather than considering `&` and `&mut` as immutable and mutable, it may make more sense to consider it as "shared" and "unique". As the Rust bindings can't guarantee that access will be unique, so `&` is more a more accurate modelling.
+3. Why godot-rust does not use RefCell (or some other form of interior mutability) is because the types already have interior mutability as they exist in Godot (and can't be tracked by Rust).
+
+## Why is there so much `unsafe` in godot-rust?
+
+Short Answer: Because all code in Godot is in C++ which cannot be statically analyzed by the compiler to guarantee safety.
+
+Longer Answer: Most of the use of the `unsafe` keyword is related to working with the scene tree as well as working with the Godot libraries. This is due to the lack of static analysis over the engine to ensure safety while still being perfectly valid code. Usage of `unsafe` does not necessarily mean that it is dangerous to use, but instead that you lose 5 protective measures in unsafe Rust as [per the nomicon](https://doc.rust-lang.org/nomicon/what-unsafe-does.html)
```
Bromeon: To me, the longer answer seems a bit abstract and doesn't really answer more than the short one. What would be interesting to mention is that a lot of `unsafe` exists primarily because of:
* **Object lifetimes:** Godot manages memory of objects independently of Rust. This means that Rust references can be invalidated when Godot destroys referred-to objects. This usually happens due to bugs in GDScript code, like calling `free()`/`queue_free()` of something actively in use.
* **Thread safety:** while Rust has a type system to ensure thread safety statically (`Send`, `Sync`), such mechanisms do not exist in either GDScript or C++. Even user-defined GDScript code has direct access to [`Thread`](https://docs.godotengine.org/en/stable/classes/class_thread.html) API.
* _other reasons?_

You can copy-paste this as-is, if you want.

src/faq/code-faq.md
```diff
@@ -0,0 +1,312 @@
...
+Longer Answer: Most of the use of the `unsafe` keyword is related to working with the scene tree as well as working with the Godot libraries. This is due to the lack of static analysis over the engine to ensure safety while still being perfectly valid code. Usage of `unsafe` does not necessarily mean that it is dangerous to use, but instead that you lose 5 protective measures in unsafe Rust as [per the nomicon](https://doc.rust-lang.org/nomicon/what-unsafe-does.html) 
+
+Here is an example.
+
+```rust
+
+fn get_a_node(&self, owner: TRef<Node>) {
+    // This is safe because it returns an option that Rust knows how to check
+    let child = owner.get_child("foo");
+    // This is safe because
```
Bromeon: "safe because **the code panics if the expected value is not present**"

src/faq/code-faq.md
```diff
@@ -0,0 +1,312 @@
...
+    // It is the programmer's responsibility to ensure that `child` is not freed before it gets used.
+    let child: Instance<Foo> = unsafe { child.assume_safe() };
+    // This is safe because we have already asserted above that we are assuming that there should be no problem and Rust
+    // can statically analyze the safety of the functions
+    child.map_mut(|c, o| {
+        c.bar(o);
+    }).expect("this should not fail");
+    // This is unsafe because it relies on Godot for function dispatch.
+    // Rust cannot know at compile time whether child will have the method `bar` that takes no parameters.
+    // It is up to the programmer to ensure that this function call will not result in undefined behavior.
```
Bromeon: The existence of the function is not the reason why `call()` is unsafe; that wouldn't result in UB, just a Godot runtime error.

The reason why `call()` is unsafe is because it can invoke functions that directly manipulate memory (`Object.free()`, `Reference.unreference()`, ...) as well as other native code which is not safe.

src/faq/code-faq.md
```diff
@@ -0,0 +1,312 @@
...
+    unsafe {
+        child.call("bar", &[])
+    };
+}
+
+```
+
+## As a native script type needs to implement `fn new(owner: &Node) -> Self`, is it possible to pass additional arguments to `new`?
+
+Unfortunately this is currently a general limitation of GDNative (see [related issue](https://github.com/godotengine/godot/issues/23260)).
```
Bromeon: I would start with "No" to make it very clear 🙂

src/faq/code-faq.md
```diff
@@ -0,0 +1,312 @@
...
+So instead of `Enemy.new()` you can write `EntityFactory.enemy(args)` in GDScript.    
+This still needs an extra type `EntityFactory`, however you could reuse that for multiple classes.
+
+
+## How can I implement static methods in GDNative?
+
+In GDScript, classes can have static methods.
+However, due to a limitation of GDNative, static methods are not supported in general. 
+
+This is another limitation of GDNative -- static methods are not supported in general.
```
Bromeon: Duplicate sentence

src/faq/code-faq.md
```diff
@@ -0,0 +1,312 @@
...
+```
+
+## Is it possible to set subproperties of a Godot type, such as a Material?
+
+Yes, it is possible to set this, but it will depend on the type.
+
+For example, with `SpatialMaterial`(https://docs.rs/gdnative/latest/gdnative/api/struct.SpatialMaterial.html) when you need to set an albedo_texture onto the material, it will be necessary to use the generic [`set_texture`](https://docs.rs/gdnative/0.9.3/gdnative/api/struct.SpatialMaterial.html#method.set_texture) function with the parameter index.
+
+While a similar case applies for [`ShaderMaterial`](https://docs.rs/gdnative/latest/gdnative/api/shader_material/struct.ShaderMaterial.html) in the case of shader material, to set the shader parameters, you will need to use the [set_param](https://docs.rs/gdnative/latest/gdnative/api/shader_material/struct.ShaderMaterial.html#method.set_shader_param) method with the relevant parameter name and value.
+
```
Bromeon: Maybe add:

> Direct access to such properties is planned in https://github.com/godot-rust/godot-rust/issues/689.

src/faq/code-faq.md
```diff
@@ -0,0 +1,312 @@
...
+## What is the godot-rust equivalent of `preload`?
+
+Unfortunately, there is no equivalent to preload in languages other than GDScript, because preload is GDScript-specific magic that works at compile time. If you read the official documentation on preload, it says:
+
+    Returns a resource from the filesystem that is loaded during script parsing. (emphasis mine)
```
Bromeon: I miss the emphasis 🙂 

Also, would it make sense to have the whole sentence as italics (to suggest quote) and the emphasis additionally in bold?

src/faq/code-faq.md
```diff
@@ -0,0 +1,312 @@
...
+You can use ResourcePreloader as a separate node in your scene, which will work regardless of whether you're using Rust or GDScript. However, note that if you create a ResourcePreloader in your code, you will still be loading these resources at the time of construction, because there is no way for the engine to know what resources are being added before actually running the code.
+
+The ResourceLoader should be used in most cases.
+
+Also, you can always put a Mutex<HashMap> into a Rust static and load everything you need there during a loading screen.
+
+## How do I keep a reference of `Node`?
+
+For example, in the following GDScript code
+```swift
```
Bromeon: Does Swift have similar syntax highlighting? Or do we want to add ` ```gdscript ` and hope that `mdbook` will support it one day?

GitHub already supports it, which is very nice:
```gdscript
extends Node
class_name MyClass
var node

func _ready():
  node = Node.new()
  self.add_child(node, false)
```

src/faq/code-faq.md
```diff
@@ -0,0 +1,312 @@
...
+  self.add_child(node, false)
+
+```
+
+```rust
+#[derive(NativeClass)]
+#[inherit(Node)]
+#[no_constructor]
+struct MyNode {
+    // It should also be possible to create an null `Ref<Node>` but
```
Bromeon: I'm actually not sure if null refs are possible, or already checked at the marshalling layer.
If we did `my_ref.assume_safe()` and then dereference the `TRef`, things would explode.

As such, I would directly suggest the `Option<Ref<Node>>`.

src/faq/configuration-faq.md
```diff
@@ -0,0 +1,136 @@
...
+[entry]
+X11.64="res://path/to/lib{name_of_binary}.so"
+OSX.64="res://path/to/lib{name_of_binary}.dylib"
+Windows.64="res://path/to/{name_of_binary}.dll"
+
+[dependencies]
+X11.64=[  ]
+OSX.64=[  ]
+Windows.64 = [ ]
+[general]
```
Bromeon: `[general]` should probably be one line further down, to make clear it belongs to the config options below.

src/faq/configuration-faq.md
```diff
@@ -0,0 +1,136 @@
...
+## How do I create the library file for my GDNative binary?
+
+You can create .gdnlib files with either of the following methods.
+
+### Editor 
+1. In the editorFrom the editor, right click in the File System and Select "Create New Resource" 
+2. Select gdnative library from the menu.
+3. Save as new gdnative library
+4. Select the gdnative library.
+5. For each desired target, select a path to the location of the library created by cargo.
```
Bromeon: Few things:
* Missing colon: "In the editor: from the editor, ..."
* In my version the point is called `New Resource...`, not `Create New Resource`, is this different for you?
* Please use `GDNativeLibrary` everywhere, I think different capitalization/spacing can confuse people here
* Maybe in point 3 we should mention: "The typical file extension for libraries is .gdnlib." -- It is referred to further down in the FAQ.

src/faq/configuration-faq.md
```diff
@@ -0,0 +1,136 @@
...
+symbol_prefix="godot_"
+reloadable=true
+```
+
+## Once I create the .gdnlib file, how do I create the GDNative Script so that I can attach it to the nodes in the scene tree?
+
+Script files can be created in two ways.
+
+1. From the editor
+2. Created manually by create a .gdns file with the following code snippet
```
Bromeon: 2. Manually by **creating** a .gdns file with the following code snippet:

src/faq/configuration-faq.md
```diff
@@ -0,0 +1,136 @@
...
+class_name = "{class name}"
+library = ExtResource( 1 )
+
+```
+
+## Tool scripts and tool classes
+
+**Question**
+
+Can I use rust as a toolscript?
```
Bromeon: "Rust"

Maybe also link to the Godot page:

> Can I use Rust as a [tool script](https://docs.godotengine.org/en/stable/tutorials/misc/running_code_in_the_editor.html)?

src/faq/configuration-faq.md
```diff
@@ -0,0 +1,136 @@
...
+
+**Answer**
+
+Yes, any rust struct that inherits from `NativeClass` can be also used as a tool class by using the `InitHandle::add_tool_class`
+
+
+For example, in the case where a
```
Bromeon: incomplete

src/faq/configuration-faq.md
```diff
@@ -0,0 +1,136 @@
...
+fn init(handle: InitHandle) {
+    handle.add_tool_class::<InitHandle>();
+}
+```
+In your init function, instead of calling 
+
+> Important considerations:
+> Tool classes that are used in the editor can cause instability when the dynamnic library is recompiled. It is important that you keep a copy of the file.
+> 
+> If your tool classes inherit `EditorPlugin`, `EditorImportPlugin` or other extensions, it is required that you set `reloadable = false` in the gdnative library where the tool class is defined.
```
Bromeon: Incomplete sentence; not really sure where the "considerations" belong or why they're formatted like this?

src/faq/configuration-faq.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Is it possible to use multiple libraries with GDNative?
+
+Yes. You can create as many libraries as you need for your project. Do be mindful that it is possible for name collision to occur when loading from multiple dynamic libraries. This can occur in the event that multiple libraries attempt to register the same class in their init function.
```
Bromeon: Good that you explain this, it's a little-known feature. Although while possible, there's usually no good reason to use multiple libraries. I would also explain that the recommended way is to have 1 .gdnlib and many .gdns files.

src/faq/help-faq.md
```diff
@@ -0,0 +1,30 @@
+# Help FAQ
```
Bromeon: Maybe name this chapter "Community"? It's not only about help but also about interacting with the community and contributing.

src/faq/help-faq.md
```diff
@@ -0,0 +1,30 @@
...
+## I need help, where can I ask?
+
+The godot-rust project uses several different sources for different kinds of information.
+
+1. The godot rust documentation is hosted at [docs.rs](https://docs.rs/gdnative/)
+
+Specific places to start would be the follow:
+- [API documentation](https://docs.rs/gdnative/0.9.3/gdnative/api/index.html) this is useful to find the necessary includes from the Godot API that are not directly implemented in the preludes.
+- [Godot Object](https://docs.rs/gdnative/0.9.3/gdnative/trait.GodotObject.html) includes all of the information about the trait that all Godot API objects implement. 
+- [NativeClass](https://docs.rs/gdnative/0.9.3/gdnative/derive.NativeClass.html) includes all the derive macro attributes that you can use to integrate your script class with godot.
```
Bromeon: * "the following:"
* Also mention the book, I would even do so as first bullet point.
* Can you replace `0.9.3` with `latest`? Then the FAQ doesn't become outdated as we release new versions.
* Maybe `GodotObject` and `NativeClass` as sub-bulletpoints of `API documentation` (i.e. indented one level).

src/faq/meta-faq.md
```diff
@@ -0,0 +1,28 @@
+# About godot-rust project FAQ
+
+## What does godot-rust's version mean?
+
+godot-rust follows the [semantic versioning](https://semver.org/) for the API breaking changes.
```
Bromeon: Rust semver is slightly different, I would link to this instead: https://doc.rust-lang.org/cargo/reference/semver.html

src/faq/meta-faq.md
```diff
@@ -0,0 +1,28 @@
...
+**Similar Question**
+
+Why is X feature not available in the API?
+Why does X function take Y parameters and return Z?
+
+**Answer**
+
+The short answer is "that is how the Godot API works". As godot-rust is focused on the bindings to Godot, we follow the API.
+
+The longer answer is that due the goal of the bindings is to ease the friction between using Rust and interacting with the Godot API. This is accomplished by use of Procedural Macros as well as clever use of traits and trait conversions to keep friction and `unsafe` blocks to a minimum.
```
Bromeon: "The longer answer is that ~due~ the goal"

src/faq/meta-faq.md
```diff
@@ -0,0 +1,28 @@
...
+
+In the case of primitive types, to avoid the runtime overhead of the FFI effort has been made to write conversions between a select number of Godot types to have a Rust native type, but in most cases, all classes and functionality is on the Godot side of the API bounds.
```
Bromeon: This sentence is quite long and still rather vague -- I would instead mention something like:

> Types, which are not part of Godot's `Object` class hierarchy, have an implementation in Rust. This avoids FFI overhead and allows code to be more idiomatic. Examples are `Vector2`, `Vector3`, `Color`, `GodotString`, `Dictionary`, etc.

src/faq/multithreading-faq.md
```diff
@@ -0,0 +1,41 @@
+# Multithreading FAQ
+
+## How do I use multithreading?
+
+Follow godot's thread safety guidelines and change your settings to "Multi-threaded" in "Project Settings > Rendering > Threading".
```
Bromeon: "Godot's thread safety guidelines" could link to https://docs.godotengine.org/en/stable/tutorials/threads/index.html.

src/faq/multithreading-faq.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Why does my game freeze while using multiple threads?
+
+Make sure you have multi-threading enabled under "Project Settings > Rendering > Threading".
+
+In addition, you will need to ensure that you are not ending up with a deadlock scenario within your Rust code.
+
+## Why is accessing the servers using multiple threads is slow?
+
+There are a few potential things that can slow down server usage
```
Bromeon: full-stop or colon

src/faq/multithreading-faq.md
```diff
@@ -0,0 +1,41 @@
...
+
+### Potential Cause #1 - Command queue is too small
+
+For multi-threaded access the servers use a thread-safe command queue. This allows you to send multiple commands from any thread. These queues have a fixed size that can be set in the threading section of the Memory -> Limits section of the Project settings.
+
+### Potential Cause #2 - Using rayon parallel iterators
+
+The reference created by a server method such as  `VisualServer::godot_singleton()` is not thread-safe. As such, you have to individually get a reference to the singleton and this can cause some severe slowdown that would not occur if the same job were run from a single thread.
+
+Multi-threading with Servers would require that you manually create a threadpool that can hold the reference to the Server. Additional testing is required to ensure that this is an actual optimization.
```
Bromeon: This explanation is absolutely awesome 👍🏼 
Very valuable insights that would need a long time to figure out.

src/faq/multithreading-faq.md
```diff
@@ -0,0 +1,41 @@
...
+
+
+## Why do I get DifferentThread when calling my code from multiple threads?
```
Bromeon: maybe "`DifferentThread` error"

src/faq/tooling-faq.md
```diff
@@ -0,0 +1,47 @@
+# Tooling
+
+## Which platforms are supported?
+
+Short Answer: Wherever you can get your code (and Godot) to run. :)
+
+Long Answer: Rust and Godot are natively available for Windows, PC, Linux, Android and iOS export targets.
+
+As of writing, WASM targets are a tenatively doable, but it requires additional configuration and is currently not supported out of the box. This godot-rust bindings do not officially support this target.
```
Bromeon: "tentatively" and "The godot-rust bindings"

I would add:
> The progress in this regard can be tracked in https://github.com/godot-rust/godot-rust/issues/647.

src/faq/tooling-faq.md
```diff
@@ -0,0 +1,47 @@
...
+
+The official position of the godot-rust project is that, we do not have the ability to offer console support due to a lacking access to the SDK.
+
+As for whether or not it is possible to compile Rust to run on a console, due to the Non-disclosure Agreements that are a part of the console SDKs, it is unlikely that anyone can give you a definitive answer.
+
+The official Godot documentation goes into more details with how this works from a Godot perspective. Please keep in mind that this does not necessarily extend to GDNative. [Console Support with Godot](https://docs.godotengine.org/en/stable/tutorials/platform/consoles.html)
+
+Regarding whether Rust can run on a console or not, as most modern consoles are using x86 processor architectures, it should be possible to compile for them. The primary issue will be whether the console manufacturer has a version of the Rust standard library that is building for their console. If not, it would be necessary to port it or leverage a port from another licensed developer.
+
+##  What about performance? Is this binding speed up final Godot build?
```
Bromeon: Maybe "Does this binding speed up my game in Godot"?

src/faq/tooling-faq.md
```diff
@@ -0,0 +1,47 @@
...
+
+Short Answer: It depends on what you are building and the bottlenecks, but it's pretty fast(tm).
+
+Longer Answer: Generally speaking, compared to GDScript, Rust will (in most cases) be a large performance increase.
+
+Compared to an C or C++ GDNative, the performance should be similar.
+
+Compared to a C++ Engine module, the performance will probably be somewhat slower as GDNative modules cannot take advantage of certain optimizations and need to use the C Foreign Function Interface at runtime.
+
+Caveat: The above apply only to release builds with appropriate release settings. When comparing to debug builds in Rust are very, very slow.
```
Bromeon: "Debug builds in Rust can be very slow".

I think it also depends what one is doing, so I would write it less absolute.

src/faq/tooling-faq.md
```diff
@@ -0,0 +1,47 @@
...
+
+Caveat 2: It is still possible to write rust code that has poor performance. You will still need to be mindful of the specific algorithms that you are choosing.
```
Bromeon: > Caveat 3: A lot of code in a game is not performance-critical and as such, Rust will help only in limited ways. When deciding between GDScript and Rust, you should also consider other factors such as static type safety, scalability, ecosystem and prototyping speed. Try to use the best of both worlds, it's very well possible to have part of your code in GDScript and part in Rust.

Bromeon (commented):
Thank you very much, your changes add some extremely valuable insights 👍🏼 

I tried to be quite open about my thoughts, so expect lots of comments 🙂

In cases where I added code or text, it can make sense if you use the "Quote comment" function to get the original formatting -- this is especially true for links within this repo, which are sugared by GitHub. Also feel free to copy-paste anything verbatim.

Some general consistency remarks: maybe check the capitalization of a few keywords -- `Rust`, `Godot`, `godot-rust`, `GDNative`, `GDScript`. At some point in the distant future I might add CI checks for those 😛

src/faq/code-faq.md
jacobsky: Oops! Good catch!

src/faq/code-faq.md
jacobsky: I like that phrasing much better, I'll go ahead and put that in the faq.

src/faq/code-faq.md
jacobsky: That's a good point, I think you are right. When I look at the example, that's the same line of reasoning that I was going for, but I couldn't put it into words directly above.

I will update it.

src/faq/code-faq.md
jacobsky: That's a much better point, I will add that.

src/faq/code-faq.md
jacobsky: Oops! I just copy pasted that from one of the questions directly. I'll add in the emphasis. :)

src/faq/code-faq.md
jacobsky: I've used swift since it gives *some* syntax highlighting. It's only marginally more accurate than regular text though.
I'll switch it over to gdscript 👍 I didn't know that github supports it 😆

src/faq/code-faq.md
jacobsky: That's a good point. I'll go ahead and remove that.

src/faq/configuration-faq.md
jacobsky: Those are all a few oversights. I'll go ahead and correct them.

src/faq/configuration-faq.md
jacobsky: Maybe just `Important` would suffice? I was trying to add some information here to let users know that using `reloadable=true` for editor plugins and that it may also be useful if the editor plugin needs to use a NativeClass script to maintain multiple libraries. I'll give this a rewrite and see if I can make it make more sense.

src/faq/configuration-faq.md
jacobsky: I'll add that in. It's definitely useful for certain use-cases, but I agree. We should probably indicate the "preferred" way of working with the library.

src/faq/help-faq.md
jacobsky: I like it! I'll change it.

src/faq/help-faq.md
jacobsky: Thank you for the feedback, I made all the changes.

src/faq/multithreading-faq.md
jacobsky: I spent quite a bit working out these errors in my own project. 😆 I hope it helps someone else 😄

src/faq/tooling-faq.md
jacobsky: I like it!

src/faq/tooling-faq.md
jacobsky: That's a good point. I'll add that.

src/faq/tooling-faq.md
jacobsky: That's a great addition, I'll add that.

jacobsky:
@Bromeon Thank you for all the valuable insight. I've gone ahead and incorporated your changes. I appreciate the thorough look into the full FAQ. If you have any additional changes, please feel free to let me know.

src/faq/configuration-faq.md
Bromeon: Wasn't about the word "considerations", rather the 2 sentences below.

I see; do you want to format them like regular bullet points (instead of `>`)?
The "Important considerations" itself should not be indented.

jacobsky: I think that might make more sense. I'll fix the formatting and push a follow-up change.

jacobsky:
@Bromeon Alrighty. I think the most recent push should resolve the last of the changes. Please let me know if there's anything else that needs doing :)

jacobsky:
Also I just noticed, but I believe this resolves issue #23 and possibly #5

Bromeon:
It also overlaps somewhat with https://github.com/godot-rust/book/pull/24.
That would be 4-in-1, do you think we'll manage 5? 😬

jacobsky:
@Bromeon To ensure that we can close #5 I went ahead and added a section on profiling to "code-faq.md" at the bottom. I'm starting to think that I'll need to work on another PR to make some sub-sections for "code-faq" since it's getting pretty long and a little mixed up/hard to follow.

src/SUMMARY.md
```diff
@@ -13,11 +13,18 @@
     - [Exported properties](./rust-binding/properties.md)
     - [Calling into GDScript from Rust](./rust-binding/calling-gdscript.md)
 - [FAQ](./faq.md)
+  - [Code Questions](./faq/code-faq.md)
+  - [Multithreading](./faq/multithreading-faq.md)
+  - [Configuration](./faq/configuration-faq.md)
+  - [Tooling](./faq/tooling-faq.md)
+  - [About the Project](./faq/meta-faq.md)
+  - [Community](./faq/community-faq.md)
```
Bromeon: The `-faq` suffix seems redundant to me here as `/faq` is already in the path. Is there a specific reason for this naming? We don't have `logging-recipe.md` either, for example.

src/faq/code-faq.md
```diff
@@ -0,0 +1,365 @@
...
+    #[export]
+    fn get_data(&self, owner: TRef<Node>) -> &LargeData {
+        &self.data
+    }
+}
+```
+
+The assumption with the above code is that `Signaller` is holding data that is too large to be feasible to clone into the signal. So the purpose of the signal is to notify other Nodes or Objects that this the data has been updated.
+
+The problem is that unless the connected node's all connect with the `Object::CONNECT_DEFERRED` it will immediately (as if it were a callback) notify the nodes and they will attempt to borrow the data. This is the root cause of the `BorrowFailed` error.
```
Bromeon: "node's" -> "nodes"

I would restructure the sentence a bit, to make it easier to understand:

> The problem is that, unless the nodes all connect with the `Object::CONNECT_DEFERRED` flag, they will be notified immediately and will attempt to borrow the data. This is the root cause of the `BorrowFailed` error.

src/faq/code-faq.md
```diff
@@ -0,0 +1,365 @@
...
+
+There are two ways to solve it.
+
+1. Ensure that all nodes use `Object::CONNECT_DEFERRED` since this will ensure that the callbacks wait until the idle_frame signal to borrow the data.
+2. Make `data` in a `RefCell<LargeData>` if it should only be accessed from the same thread (such as with signals) or `Mutex<LargeData>` if you need thread-safety to be important. Then you can modify update date to the following
```
Bromeon: maybe "Store `data`" instead of "Make"?

"to be important" can be omitted, I think.

Last sentence is incomplete, and I think it should be "`update_data()`".

src/faq/code-faq.md
```diff
@@ -0,0 +1,365 @@
...
+    }
+```
+
+In both instances you will not encounter the reentrant errors.
+
+## Why are methods on all Godot API classes using `&self` even when they should cause mutations?
+
+1. `&mut` means that only one reference can exist simultaneously (no aliasing), _not_ that the object is mutable. Mutability is often a _consequence_ of the exclusiveness. Since Godot objects are managed by the engine, Rust cannot guarantee that references are exclusive, as such using `&mut` would cause undefined behavior. Instead, an interior mutability pattern based on `&T` is used. 
+  For more information, please refer to [this explanation](https://docs.rs/dtolnay/latest/dtolnay/macro._02__reference_types.html) for more information.
+2. In godot-rust, rather than considering `&` and `&mut` as immutable and mutable, it may make more sense to consider it as "shared" and "unique". As the Rust bindings can't guarantee that access will be unique, so `&` is more a more accurate modelling.
```
Bromeon: Now that I read it again, point 2 and 1 are exactly the same. I think we can remove point 2 entirely, no?

src/faq/code-faq.md
```diff
@@ -0,0 +1,365 @@
...
+
+## Why is there so much `unsafe` in godot-rust?
+
+Short Answer: Because all code in Godot is in C++ which cannot be statically analyzed by the compiler to guarantee safety.
+
+Longer Answer: The reasons why `unsafe` is required is that 
+
+- Object lifetimes: Godot manages memory of objects independently of Rust. This means that Rust references can be invalidated when Godot destroys referred-to objects. This usually happens due to bugs in GDScript code, like calling free()/queue_free() of something actively in use.
+- Thread safety: while Rust has a type system to ensure thread safety statically (Send, Sync), such mechanisms do not exist in either GDScript or C++. Even user-defined GDScript code has direct access to Thread API.
+- C FFI: Any interactions that cross the C Foreign Function Interface will be unsafe by default as Rust cannot guarantee safety. While many functions may be safely reasoned about, there are still some functions which will be inherently unsafe due to their potential effects on object lifetimes.
```
Bromeon: Good point! 👍🏼

src/faq/code-faq.md
```diff
@@ -0,0 +1,365 @@
...
+
+One of the ways that godot-rust avoids 
+
+Here is a documented example with some of the unsafe usage that you will seem
```
Bromeon: Both incomplete 🙂

src/faq/code-faq.md
```diff
@@ -0,0 +1,365 @@
...
+The type needs to be instantiated somewhere on GDScript level.
+Good places for instantiation are for instance:
+
+- as a member of a long-living util object,
+- as a [singleton auto-load object](https://docs.godotengine.org/en/stable/getting_started/step_by_step/singletons_autoload.html).
+
+
+## How to I convert from a `Variant` or other Godot Type to the underlying Rust type?
+
+Assuming that a method takes an argument `my_object` as a `Variant`
```
Bromeon: I think it's not clear what `my_object` is. Maybe `Object` is also not a typical example, because people usually deal with nodes.

Should we assume that the declaration looks something like this:
```rs
#[derive(NativeClass)]
#[inherit(Node2D)] // something specific, so it's clear when this type re-occurs in code
struct MyNode { ... } // instead of MyObject
```

We can also do the same thing with `Object` instead of `Node2D`, but then `self` (i.e. `AnotherNativeScript`) should not _also_ inherit `Object`.

src/faq/code-faq.md
```diff
@@ -0,0 +1,365 @@
...
+        let my_object = unsafe {
+            my_object
+                .try_to_object::<Object>()
+                .expect("Failed to convert my_object variant to object")
+                .assume_safe()
+        };
+        // 2. Obtain a RefInstance.
+        let my_object = my_object
+            .cast_instance::<MyObject>()
+            .expect("Failed to cast my_object object to instance");
```
Bromeon: Mention that step 2 gives access to the Rust object (i.e. custom user data).

src/faq/code-faq.md
```diff
@@ -0,0 +1,365 @@
...
+
+```
+
+## Is it possible to set subproperties of a Godot type, such as a Material?
+
+Yes, it is possible to set this, but it will depend on the type.
+
+For example, with `SpatialMaterial`(https://docs.rs/gdnative/latest/gdnative/api/struct.SpatialMaterial.html) when you need to set an albedo_texture onto the material, it will be necessary to use the generic [`set_texture`](https://docs.rs/gdnative/latest/gdnative/api/struct.SpatialMaterial.html#method.set_texture) function with the parameter index.
+
+While a similar case applies for [`ShaderMaterial`](https://docs.rs/gdnative/latest/gdnative/api/shader_material/struct.ShaderMaterial.html) in the case of shader material, to set the shader parameters, you will need to use the [set_param](https://docs.rs/gdnative/latest/gdnative/api/shader_material/struct.ShaderMaterial.html#method.set_shader_param) method with the relevant parameter name and value.
```
Bromeon: `SpatialMaterial` link broken, missing []

Can you add backticks around `albedo_texture` and `set_param`?

src/faq/code-faq.md
```diff
@@ -0,0 +1,365 @@
...
+
+The GDNative API supports any type that implements the [`ToVariant`](https://docs.rs/gdnative/latest/gdnative/core_types/trait.ToVariant.html) and/or [`FromVariant`](https://docs.rs/gdnative/latest/gdnative/core_types/trait.FromVariant.html) traits.
+
+To use a type as a property, in addition to the above, the type will also need to implement the [`Export`](https://docs.rs/gdnative/latest/gdnative/nativescript/trait.Export.html) trait.
+
+Some concrete examples of types that can be used with the GDNative API are the following:
+
+- [`Variant`](https://docs.rs/gdnative/latest/gdnative/core_types/struct.Variant.html), this is Godot's "any" type. It must be converted before it can be used.
+- A subset of scalar types such as `i64`, `f64`, `bool`, etc.
+- [`String`] and [`GodotString`](https://docs.rs/gdnative/latest/gdnative/core_types/struct.GodotString.html).
```
Bromeon: `String` could link to https://doc.rust-lang.org/std/string/struct.String.html

src/faq/code-faq.md
```diff
@@ -0,0 +1,365 @@
...
+- [Godot Core Types](https://docs.rs/gdnative/latest/gdnative/core_types/index.html) such as [`Color`](https://docs.rs/gdnative/latest/gdnative/core_types/struct.Color.html), [`Aabb`](https://docs.rs/gdnative/latest/gdnative/core_types/struct.Aabb.html), [`Transform2D`](https://docs.rs/gdnative/latest/gdnative/core_types/type.Transform2D.html), [`Vector2`](https://docs.rs/gdnative/latest/gdnative/core_types/type.Vector2.html), etc.
+- Godot classes such as `Node`, `Reference`, etc. which must be accessed via [`Ref<T>`](https://docs.rs/gdnative/latest/gdnative/struct.Ref.html) through Ref<T> (you can't pass them by value, because Godot owns them).
```
Bromeon: remove "through Ref", we already have "via Ref"

src/faq/code-faq.md
```diff
@@ -0,0 +1,365 @@
...
+## How can I profile my code to determine the performance?
+
+There are a lot of ways to profile your code and they will vary in complexity.
+
+The simplest way to profile your code is by using the `#[gdnative::profiled]` procedural macro that enable the Godot to profile the attributed function. This option is useful for comparing performance gains when porting GDScript code to Rust or as a way to understand the relative "frame time" of your code.
+
+For more information about the Godot profiler, please refer to the [Official Documentation](https://docs.godotengine.org/en/stable/tutorials/debug/debugger_panel.html?highlight=profiler#profiler).
+
+In order for Godot to profile your function all of the following must be true:
+- The function belongs to a struct that derives `#[NativeClass]`
```
Bromeon: I would just write `NativeClass` -- `#[NativeClass]` never appears in code.

src/faq/code-faq.md
```diff
@@ -0,0 +1,365 @@
...
+- The function is included in an `impl` block that is attributed with the `#[methods]` attribute
+- The function is attributed with `#[export]` attribute
+
+As such, this method is _only_ useful for exported code and is subject
```
Bromeon: incomplete

src/faq/tooling-faq.md
```diff
@@ -0,0 +1,49 @@
+# Tooling
+
+## Which platforms are supported?
+
+Short Answer: Wherever you can get your code (and Godot) to run. :)
+
+Long Answer: Rust and Godot are natively available for Windows, PC, Linux, Android and iOS export targets.
```
Bromeon: "Windows, PC" should just be "Windows", no?

src/faq/tooling-faq.md
```diff
@@ -0,0 +1,49 @@
...
+
+As of writing, WASM targets are a tentatively doable, but it requires additional configuration and is currently not supported out of the box. The godot-rust bindings do not officially support this target. The progress in this regard can be tracked in [godot-rust/godot-rust#647](https://github.com/godot-rust/godot-rust/issues/647).
+
+## Does godot-rust support consoles?
+
+The official position of the godot-rust project is that, we do not have the ability to offer console support due to a lacking access to the SDK.
+
+As for whether or not it is possible to compile Rust to run on a console, due to the Non-disclosure Agreements that are a part of the console SDKs, it is unlikely that anyone can give you a definitive answer.
+
+The official Godot documentation goes into more details with how this works from a Godot perspective. Please keep in mind that this does not necessarily extend to GDNative. [Console Support with Godot](https://docs.godotengine.org/en/stable/tutorials/platform/consoles.html)
```
Bromeon: Maybe the link can be embedded directly into the text instead of appended at the end. Press "quote comment" to get the formatting:

The official Godot documentation [goes into more details](https://docs.godotengine.org/en/stable/tutorials/platform/consoles.html) with how this works from a Godot perspective. Please keep in mind that this does not necessarily extend to GDNative.

src/faq/tooling-faq.md
```diff
@@ -0,0 +1,49 @@
...
+
+Compared to a C++ Engine module, the performance will probably be somewhat slower as GDNative modules cannot take advantage of certain optimizations and need to use the C Foreign Function Interface at runtime.
+
+Caveat: The above apply only to release builds with appropriate release settings. When comparing to debug builds in Rust can be very slow.
+
+Caveat 2: It is still possible to write rust code that has poor performance. You will still need to be mindful of the specific algorithms that you are choosing.
+
+Caveat 3: A lot of code in a game is not performance-critical and as such, Rust will help only in limited ways. When deciding between GDScript and Rust, you should also consider other factors such as static type safety, scalability, ecosystem and prototyping speed. Try to use the best of both worlds, it's very well possible to have part of your code in GDScript and part in Rust.
+
+## How can I mke my debug builds more performant?
```
Bromeon: "make" typo

Bromeon (changes requested):
I agree, this PR is already big enough, we can always add more and refine things in the future 🙂

src/SUMMARY.md
jacobsky: No specific reason, I just named it that way for consistency initially. I think you are right, we can go ahead and remove the faq from the name.

src/faq/code-faq.md
jacobsky: Thank you for the change, I'll include it :)

src/faq/code-faq.md
jacobsky: I'll update it.

src/faq/code-faq.md
jacobsky: Yeah, that's a good point. I suppose unless we really needed to emphasize it, it would be better to go ahead and remove number 2.

src/faq/code-faq.md
jacobsky: I'll fix that up. Oops!

src/faq/code-faq.md
jacobsky: That's a much better example. I'll make the change.

src/faq/code-faq.md
jacobsky: Ah! I forgot that!

src/faq/code-faq.md
jacobsky: 😅 I'll finish that.

src/faq/tooling-faq.md
jacobsky: 😅

Bromeon:
Thanks a lot for this huge contribution! Will definitely help a lot of newcomers to navigate 🙂

PR #34, opened 2021-09-06
Added a recipe for loading godot-external recipes.

jacobsky:
I added this recipe to help users that may need to include non-Godot resources for their rust-side configuration. Please let me know if there are any additional changes or tweaks that you would like me to make to this.

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -24,6 +24,7 @@
 - [Recipes](./recipes.md)
     - [Rust Panic Hook](./recipes/rust_panic_handler.md)
     - [Logging](./recipes/logging.md)
+    - [Loading external resources](./recipes/loading_external_resources.md)
 - [Exporting](./exporting.md)
     - [Android](./exporting/android.md)
     - [(TODO) iOS](./exporting/ios.md)
--- a/src/recipes/loading_external_resources.md
+++ b/src/recipes/loading_external_resources.md
@@ -1,0 +1,96 @@
+# Loading External-Resource Files
+
+If you need to use any files that aren't explicitly supported by Godot's [Resource Loader](https://docs.godotengine.org/en/stable/classes/class_resourceloader.html), it will be necessary to process and load the file yourself.
+
+This recipe covers two methods of loading these files.
+
+## Option 1 - Embed the File into the Binary
+
+The simplest way is to embed the resource directly into your binary by using the `std::include_bytes` macro.
+
+To embed the file directly into the binary you can use the following macro:
+
+```rust
+// To have global immutable access to the file.
+const RESOURCE: &'static [u8] = include_bytes!("path/to/resource/file");
+
+fn process_the_resource() {
+    // Include the file locally
+    let bytes = include_bytes!("path/to/resource/file");
+}
+```
+
+This can be useful for embedding any information that should be included at build time.
+
+For example: such as if you wish to hard-code certain features like cheat codes, developer consoles, or default user configuration as a file rather than a build flag.
+
+This approach is much more limited as it requires recompiling for all targets whenever changes to the resources are made.
+
+## Option 2 - Embed the File in the PCK
+
+For most other use-cases you can use Godot's PCK file to export your resources into the game.
+
+This can be accomplished by using the `gdnative::api::File` module as follows.
+
+```rust
+
+fn load_resource_as_string(filepath: &str) -> String {
+    use gdnative::api::File;
+    let file = File::new();
+    file.open(filepath, File::READ).expect(&format!("{} must exist", filepath));
+    
+    let data: GodotString = file.get_as_text();
+    // Depending upon your use-case you can also use the following methods depending upon your use-case.
+    // let line: StringArray = file.get_csv_line(0);
+    // let file_len = file.get_len();
+    // let bytes: ByteArray = file.get_bytes(file_len);
+    data.to_string()
+}
+```
+
+See the [`File` Class Documentation](https://docs.rs/gdnative/latest/gdnative/api/struct.File.html) for every function that you use for loading the resources.
+
+After you retrieve the data in the desired format, you can process it like you would normal Rust code.
+
+## Option #3 Save and Load filedata as user_data
+
+Godot allows access to device side user directory for the project under "user://". This works very similar to loading above and it uses the `gdnative::api::File` API
+
+Note: Saving only works on resource paths (paths starting with "res://") when Godot is being run in the editor. After exporting "res://" becomes read-only.
+
+Example on writing and reading string data.
+
+```rust
+fn save_data_from_string(filepath: &str, data: &str) {
+    use gdnative::api::File;
+    let file = File::new();
+    file.open(filepath, File::WRITE).expect(&format!("{} must exist", &filepath));
+    
+    file.store_string(data);
+}
+
+
+fn load_data_as_string(filepath: &str) -> String {
+    use gdnative::api::File;
+    let file = File::new();
+    file.open(filepath, File::READ).expect(&format!("{} must exist", &filepath));
+    
+    let data: GodotString = file.get_as_text();
+    data.to_string()
+}
+```
+
+For more information on the paths, please refer to the [File System Tutorial](https://docs.godotengine.org/en/3.0/getting_started/step_by_step/filesystem.html#resource-path).
+
+
+### Testing
+
+This section is for unit testing from Rust without loading Godot. As `gdnative::api::File` requires that Godot be running, any Rust-only unit tests will require a separate method to be implemented in order to load the resources. This can be accomplished by creating separate code paths or functions `#[cfg(test)]` and `#[cfg(not(test))]` attributes to differentiate between test configuration and Godot library configurations.
+
+In test configurations, you will need to ensure that your loading code uses `std::fs::File` or some equivalent to read your load.
+
+### Exporting
+
+When exporting your game, under the `Resources` Tab you will need to add a filter so that godot will pack those resources into the .pck file.
+
+For example: If you are using .json, .csv and .ron files, you will need to use include `*.json, *.csv, *.ron` in the "Filters to Export non-resource files/folders" field.


src/recipes/loading_external_resources.md
```diff
@@ -0,0 +1,61 @@
...
+}
+```
+
+This can be useful for embedding any information that should be included at build time.
+
+For example: such as if you wish to hard-code certain features (e.g. cheat codes, developer consoles, default user configuration) and organize them as a file rather than a build flag.
+
+This approach is much more limited as it requires recompiling for all targets whenever changes to the resources are made.
+
+## Option 2 - Embedd the File in the PCK
```
Bromeon: "Embed" 🙂

src/recipes/loading_external_resources.md
```diff
@@ -0,0 +1,61 @@
...
+
+For most other use-cases you can use Godot's PCK file to export your resources into the game.
+
+This can be accomplished by using the `gdnative::apis::File` module as follows.
```
Bromeon: `api`, not `apis`

src/recipes/loading_external_resources.md
```diff
@@ -0,0 +1,61 @@
...
+
+```rust
+fn load_resource_as_string(filepath: &str) -> String {
+    use gdnative::api::File;
+    let file = File::new();
+    file.open(filepath, File::READ).expect(&format!("{} must exist", &filepath));
+    
+    let data: GodotString = file.get_as_text();
+    // Depending upon your use-case you can also use the following methods depending upon your use-case.
+    // let line: PoolStringAray = file.get_csv_line(0);
```
Bromeon: This type is currently called [`StringArray`](https://docs.rs/gdnative/0.9.3/gdnative/core_types/type.StringArray.html) in godot-rust, I'm not very happy with the current naming though: https://github.com/godot-rust/godot-rust/issues/773

Same for `ByteArray` below.

Bromeon (commented):
Thanks, this is very helpful!

Maybe as a third option, we should mention files that are opened normally from Rust (via `std::fs`), and live outside the .pkg file? There could be a few things to consider regarding paths. I could imagine this for anything that also needs to be written (config, savegames, highscores etc.) -- what do you think?

src/recipes/loading_external_resources.md
jacobsky: Thanks for the correction!

src/recipes/loading_external_resources.md
jacobsky: Oops! I'll fix that.

jacobsky:
> Thanks, this is very helpful!

Glad to hear it.

> Maybe as a third option, we should mention files that are opened normally from Rust (via std::fs), and live outside the .pkg file? > There could be a few things to consider regarding paths. I could imagine this for anything that also needs to be written (config,savegames, highscores etc.) -- what do you think?

I think it would be useful to mention files that live outside the `.pck` file. It would be useful to -- at the very least -- confirm that it's mostly the same. The one problem that I've found with using the `std::fs::File` is that it's difficult to get the correct directories from Godot. My first attempt at doing this was to get the root directory, but it seems like Godot intentionally obfuscates that.

For now, I'll go ahead and include a mention of saving and loading from the "user://" with the `gdnative::api::File` since it is also likely something that users will be interested in noting.

If you have any idea on how to get the canonical paths from Godot, I'd be happy to add another option for using `std::fs::File`.

Bromeon:
Ah good, wasn't aware about the troubles with getting the root path. Feel free to also leave some TODOs in italics in the FAQ, that way we won't miss that some parts still need completing.

What do you think about Rust file I/O that disregards the Godot paths (`res://` and `user://`)? It's probably more an anti-pattern, no?

jacobsky:
> Ah good, wasn't aware about the troubles with getting the root path. Feel free to also leave some TODOs in italics in the FAQ, that way we won't miss that some parts still need completing.

Will do :)

> What do you think about Rust file I/O that disregards the Godot paths (`res://` and `user://`)? It's probably more an anti-pattern, no?

From my experimentation, I'd say using Rust I/O is probably less idiomatic for a Godot project. Since Godot makes a large number of assumptions about your project structure, I'd probably say that the general case would be to use `ResourceLoader` or `File` first and only if those are not an option try to work with Rust I/O,

That being said, I doubt there's any reason that you couldn't, but it would take some additional effort to ensure that you can always find "res://" or "user://" and that you can read from the .PCK file. While .PCK is a fancy zip file, it still requires grabbing another package to read through the ZIP file (or is that supported natively in rust?) and then you'd need to do the extra leg work. It's probably just easier to use the `File` API, 😆

jacobsky:
Fixed up a merge conflict with SUMMARY.md between this and the previous commit.

src/recipes/loading_external_resources.md
```diff
@@ -0,0 +1,93 @@
...
+
+If you need to use any files that aren't explicitly supported by Godot's [Resource Loader](https://docs.godotengine.org/en/stable/classes/class_resourceloader.html), it will be necessary to process and load the file yourself.
+
+This recipe covers two methods of loading these files.
+
+## Option 1 - Embed the File into the Binary
+
+The simplest way is to embed the resource directly into your binary by using the `std::include_bytes` macro
+
+To embedd the file directly into the binary you can use
```
Bromeon: "embed"

and I would generally close sentences which refer to a following section with a colon ":".

src/recipes/loading_external_resources.md
```diff
@@ -0,0 +1,93 @@
...
+To embedd the file directly into the binary you can use 
+
+```rust
+    // To have global immutible access to the file.
+    const RESOURCE: &'static [u8] = include_bytes!("path/to/resource/file");
+fn process_the_resource() {
+    // Include the file locally
+    let bytes = include_bytes!("path/to/resource/file");
+}
+```
```
Bromeon: Indentation is strange here -- can you un-indent the `const` and comment?
Also, an empty line between the constant and the function definition would make things a bit clearer.

src/recipes/loading_external_resources.md
```diff
@@ -0,0 +1,93 @@
...
+For most other use-cases you can use Godot's PCK file to export your resources into the game.
+
+This can be accomplished by using the `gdnative::api::File` module as follows.
+
+```rust
+
+fn load_resource_as_string(filepath: &str) -> String {
+    use gdnative::api::File;
+    let file = File::new();
+    file.open(filepath, File::READ).expect(&format!("{} must exist", &filepath));
```
Bromeon: `filepath` is `&str`, so no need to take a reference to it.

src/recipes/loading_external_resources.md
```diff
@@ -0,0 +1,93 @@
...
+    
+    let data: StringArray = file.get_as_text();
```
Bromeon: `get_as_text()` returns `GodotString`, not `StringArray`.

src/recipes/loading_external_resources.md
```diff
@@ -0,0 +1,93 @@
...
+    const RESOURCE: &'static [u8] = include_bytes!("path/to/resource/file");
+fn process_the_resource() {
+    // Include the file locally
+    let bytes = include_bytes!("path/to/resource/file");
+}
+```
+
+This can be useful for embedding any information that should be included at build time.
+
+For example: such as if you wish to hard-code certain features (e.g. cheat codes, developer consoles, default user configuration) and organize them as a file rather than a build flag.
```
Bromeon: "such as" and "e.g." are redundant when you already have "For example:"

src/recipes/loading_external_resources.md
Bromeon: You resolved it, but it seems not changed, or am I missing something?

src/recipes/loading_external_resources.md
```diff
@@ -0,0 +1,93 @@
...
+    let data: StringArray = file.get_as_text();
+    // Depending upon your use-case you can also use the following methods depending upon your use-case.
+    // let line: PoolStringAray = file.get_csv_line(0);
+    // let file_len = file.get_len();
+    // let bytes: PoolByteArray = file.get_bytes(file_len);
+    data.to_string()
+}
+```
+
+See the [File Class Documentation](https://docs.godotengine.org/en/stable/classes/class_file.html) for every function that you use for loading the resources.
```
Bromeon: Should we link to https://docs.rs/gdnative/latest/gdnative/api/struct.File.html instead?
There's still a link there to the Godot doc, but I think it would be more helpful to know the involved Rust types.

Also, maybe write "`File` class documentation" 🙂

src/recipes/loading_external_resources.md
```diff
@@ -0,0 +1,93 @@
...
+
+Example on writing and reading string data.
+
+```rust
+fn save_data_from_string(filepath: &str, data: &str) {
+    use gdnative::api::File;
+    let file = File::new();
+    file.open(filepath, File::WRITE).expect(&format!("{} must exist", &filepath));
+    
+    file.store_string(data.to_owned());
```
Bromeon: [`File::store_string()`](https://docs.rs/gdnative/0.9.3/gdnative/api/struct.File.html#method.store_string) takes a parameter `string: impl Into<GodotString>`. So it can accept any `T` for which `From<T>` is implemented for `GodotString`.

[There is](https://docs.rs/gdnative/latest/gdnative/core_types/struct.GodotString.html#impl-From<S>) a
```rust
impl<S> From<S> for GodotString
   where S : AsRef<str>
```
so we should be able to pass `data` directly, no? Would be nice to avoid the intermediate copy through `String`.

src/recipes/loading_external_resources.md
```diff
@@ -0,0 +1,93 @@
...
+}
+
+
+fn load_data_as_string(filepath: &str) -> String {
+    use gdnative::api::File;
+    let file = File::new();
+    file.open(filepath, File::READ).expect(&format!("{} must exist", &filepath));
+    
+    let data: StringArray = file.get_as_text();
```
Bromeon: same here: `GodotString`

src/recipes/loading_external_resources.md
```diff
@@ -0,0 +1,93 @@
...
+    data.to_string()
+}
+```
+
+For more information on the paths, please refer to the [File System Tutorial](https://docs.godotengine.org/en/3.0/getting_started/step_by_step/filesystem.html#resource-path) for more inforamtion.
+
+
+### Testing
+
+For test configurations, it will be necessary to use `std::fs:File` to load the resources for testing. This can be accomplished by creating separate code paths or functions `#[cfg(test)]` and `#[cfg(not(test))]` attributes.
```
Bromeon: It's not entirely clear to me why this is necessary -- do you mean that one should have a separate, Rust-only configuration for tests?

Bromeon (changes requested):
Thanks for the additions!

src/SUMMARY.md
```diff
@@ -24,6 +24,7 @@
 - [Recipes](./recipes.md)
     - [Rust Panic Hook](./recipes/rust_panic_handler.md)
     - [Logging](./recipes/logging.md)
+    - [Loading External Dependencies](./recipes/loading_external_resources.md)
```
Bromeon: I would keep the name here and the title in sync. I think "Resource files" fits better than "Dependencies".

Capitalization-wise, I would only capitalize the first word and symbol names; i.e. here I'd write "Loading external resource files" or "Loading external resources" (and 1:1 for the title in the .md file).

src/recipes/loading_external_resources.md
jacobsky: I'll make sure that this gets fixed. (I may have forgotten to push the changes 😨).

src/recipes/loading_external_resources.md
jacobsky: That's a good point! I'll make the changes.

src/recipes/loading_external_resources.md
jacobsky: That's a good point. I just verified it in my other code and I do not use `to_owned()` there.

src/recipes/loading_external_resources.md
jacobsky: This section is for unit/integration testing from rust without loading Godot. As `gdnative::api::File` requires that Godot be running to use it, it's necessary for any code that is going to be loaded to have a test configuration in order to load it. I'm not sure if this section is entirely necessary, but I thought it would be useful to include. If you have any suggestions, I would be happy to make modifications.

Bromeon: I didn't question the necessity of the section, but literally the sentence "it will be necessary to use `std::fs:File`" 😉 

Definitely keep the section! I think adding what you mentioned would clarify a lot for the reader:
> This section is for unit testing from Rust without loading Godot. As `gdnative::api::File` requires that Godot be running to use it, it's necessary for any code that is going to be loaded to have a test configuration in order to load it.

jacobsky: Thanks for the input. I've made the changes.

jacobsky:
@Bromeon I made the requested changes.

src/recipes/loading_external_resources.md
```diff
@@ -0,0 +1,96 @@
...
+fn load_resource_as_string(filepath: &str) -> String {
+    use gdnative::api::File;
+    let file = File::new();
+    file.open(filepath, File::READ).expect(&format!("{} must exist", filepath));
+    
+    let data: GodotString = file.get_as_text();
+    // Depending upon your use-case you can also use the following methods depending upon your use-case.
+    // let line: PoolStringArray = file.get_csv_line(0);
+    // let file_len = file.get_len();
+    // let bytes: PoolByteArray = file.get_bytes(file_len);
```
Bromeon: `StringArray/ByteArray` 🙂

src/recipes/loading_external_resources.md
```diff
@@ -0,0 +1,96 @@
...
+    use gdnative::api::File;
+    let file = File::new();
+    file.open(filepath, File::READ).expect(&format!("{} must exist", &filepath));
+    
+    let data: GodotString = file.get_as_text();
+    data.to_string()
+}
+```
+
+For more information on the paths, please refer to the [File System Tutorial](https://docs.godotengine.org/en/3.0/getting_started/step_by_step/filesystem.html#resource-path) for more inforamtion.
```
Bromeon: "inforamtion" typo

src/recipes/loading_external_resources.md
```diff
@@ -0,0 +1,96 @@
...
+
+
+### Testing
+
+This section is for unit testing from Rust without loading Godot. As `gdnative::api::File` requires that Godot be running to use it, it's necessary for any code that is going to be loaded to have a test configuration in order to load it. This can be accomplished by creating separate code paths or functions `#[cfg(test)]` and `#[cfg(not(test))]` attributes to differentiate between test configuration and Godot library configurations.
```
Bromeon: > As `gdnative::api::File` requires that Godot be running to use it, it's necessary for any code that is going to be loaded to have a test configuration in order to load it.

Sounds a bit like circular reasoning; I would probably write something like:

As `gdnative::api::File` requires that Godot be running, a Rust-only unit test for loading needs to be implemented without it.

src/recipes/loading_external_resources.md
```diff
@@ -0,0 +1,96 @@
...
+This recipe covers two methods of loading these files.
+
+## Option 1 - Embed the File into the Binary
+
+The simplest way is to embed the resource directly into your binary by using the `std::include_bytes` macro.
+
+To embed the file directly into the binary you can use the following macro:
+
+```rust
+// To have global immutible access to the file.
```
Bromeon: "immutable"

src/recipes/loading_external_resources.md
jacobsky: Ah, right they are named differently in gdnative. Ooops!

jacobsky:
@Bromeon I just finished the last round of changes. Please let me know if there are any additional. If not, I'll go ahead and squash the commits for merging.

PR #35, opened 2021-09-11
Added FAQ Info about adding custom types to the editor via a plugin.

jacobsky:
I wanted to capture this information that we were discussing in discord. I think this may be useful to many folks that make more heavy use of the editor.

Please let me know if you think there are any changes that are necessary.

### DIFF
--- a/src/faq/configuration.md
+++ b/src/faq/configuration.md
@@ -51,6 +51,12 @@ library = ExtResource( 1 )
 
 ```
 
+## Why aren't my scripts showing up in the Add Node portion of the editor even though they inherit from  `Node`, `Node2D`, `Spatial` or `Control`?
+
+Due to limitations with Godot 3.x's version of GDNative, NativeScript types do not get registered in the class database. As such, these classes are not included in the "Create New Node" dialog.
+
+A workaround to this issue has been included in the recipe [Custom Nodes Plugin](../recipes/custom-nodes-plugin.md).
+
 ## Can I use Rust for a [toolscript](https://docs.godotengine.org/en/stable/tutorials/misc/running_code_in_the_editor.html)?
 
 Yes, any Rust struct that inherits from `NativeClass` can be also used as a tool class by using the `InitHandle::add_tool_class` during native script initialization instead of `InitHandle::add_class`.
--- a/src/recipes/custom-nodes-plugin.md
+++ b/src/recipes/custom-nodes-plugin.md
@@ -1,0 +1,51 @@
+# Custom Nodes Plugin
+
+While GDNative in Godot 3.x has some limitations, creating an `EditorPlugin` that registers custom nodes can be used to help integrate your types more tightly in the engine. Using a Custom Nodes plugin, you may register as many `NativeClass` scripts as you need in a single plugin.
+
+1. Create the directory "res://addons/my_custom_nodes/".
+2. In the directory "res://addons/my_custom_nodes/", create `plugin.cfg` and `add_my_custom_nodes.gd` and include the following code for each.
+
+## plugin.cfg
+```ini
+[plugin]
+name="My Custom Nodes"
+description="Adds my custom nodes for my native classes."
+author="your-name"
+version="0.1.0"
+script="add_my_custom_nodes.gd"
+```
+
+## add_my_custom_nodes.gd
+```gdscript
+tool
+extends EditorPlugin
+
+func _enter_tree():
+    # Get the base icon from the current editor interface/theme
+    var gui = get_editor_interface().get_base_control()
+    var node_icon = gui.get_icon("Node", "EditorIcons")
+    
+    # Alternatively, preload to a custom icon at the following
+    # var node_icon preload("res://icon_ferris.png")
+
+    add_custom_type(
+        "MyNode",
+        "Control",
+        preload("res://gdnative/MyNode.gdns"),
+        node_icon
+    )
+
+    # Add any additional custom nodes here here.
+
+func _exit_tree():
+    remove_custom_type("MyNode")
+    # Add a remove for each registered custom type to clean up
+```
+
+From the editor's main menu, find "Project > Project Settings > Plugins". Find "My Custom Nodes" in the list of plugins and activate this. From here, you can create these directly from the "Create New Node" dialog. In addition, any nodes created this way will disallow the script from being changed or removed.
+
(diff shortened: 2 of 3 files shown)


src/faq/configuration.md
```diff
@@ -51,6 +51,58 @@ library = ExtResource( 1 )
 
 ```
 
+## Why aren't my scripts showing up in the Add Node portion of the editor even though they inherit from  `Node`, `Node2D`, `Spatial` or `Control`?
+
+Due to limitations with the current version of GDNative, the Native Script types do not get registered in the class database. This is used by Godot to populate the classes registered do not get added to the class database which is requiredclass database does not get updated with GDNative scripts.
+
```
Bromeon: The sentence has gone through the grinder 😀

src/faq/configuration.md
```diff
@@ -51,6 +51,58 @@ library = ExtResource( 1 )
...
+A workaround to this issue that will let your nodes be added to the editor interface is to create an editor plugin. Such as the following example.
```
Bromeon: I wonder if the editor plugin could make sense as a standalone recipe?
We can also leave it here for now, up to you.

src/faq/configuration.md
```diff
@@ -51,6 +51,58 @@ library = ExtResource( 1 )
...
+	# Get the base icon from the current editor interface/theme
+	var gui = get_editor_interface().get_base_control()
+	var node_icon = gui.get_icon("Node", "EditorIcons")
+	add_custom_type(
+		"MyNode",
+		"Control",
+		preload("res://gdnative/MyNode.gdns"),
+		node_icon
+		# Alternatively, preload to a custom icon at the following
+#		preload("res://icon_ferris.png")
```
Bromeon: I would probably add the commented-out parts immediately in the lines after `node_icon`, it's a bit too busy in the argument list.

Maybe the `add_custom_type` call can also be after an empty line, to clarify that this is the actual call to register the type and separate it from the "preparation work".

src/faq/configuration.md
```diff
@@ -51,6 +51,58 @@ library = ExtResource( 1 )
...
+# plugin.cfg
+[plugin]
+name="My Custom Nodes"
+description="Adds my custom nodes for my native classes."
+author="your-name"
+version="0.1.0"
+script="add_my_custom_nodes.gd"
+```
+
+```gdscript
```
Bromeon: This snippet has some mixed tabs and spaces, would it be possible to unify it to 4 spaces per indentation level?

src/faq/configuration.md
```diff
@@ -51,6 +51,58 @@ library = ExtResource( 1 )
...
+func _exit_tree():
+	remove_custom_type("MyNode")
+    # Add a remove for each registered custom type to clean up
+```
+
+From the editor's main menu, find "Project > Project Settings > Plugins". Find "My Custom Nodes" in the list of plugins and activate this. From here, you can create these directly from the "Create New Node" dialog. In addition, any nodes created this way will disallow the script from being changed or removed.
+
+For more information about this kind of plugin, please refer to the [Godot Documentation](https://docs.godotengine.org/en/stable/tutorials/plugins/editor/making_plugins.html#a-custom-node).
+
+**Note**: This method only works for script types and changing the path to the ".gdns" file will cause the custom `Node` to no longer register correctly.
```
Bromeon: what do you mean with "script types" in this context?

src/faq/configuration.md
```diff
@@ -51,6 +51,58 @@ library = ExtResource( 1 )
...
+
+**Note 2**: This does not register it in the class database, in order to use the script through GDScript or other Godot based means, it will still be necessary to reference the script path directly.
```
Bromeon: maybe replace "it" with `add_my_custom_nodes` -- or what class are you referring to?

Also, maybe full-stop before "in order"?

Bromeon (changes requested):
Thank you! 👍🏼

src/faq/configuration.md
jacobsky: I'll rework it.

src/faq/configuration.md
jacobsky: I hadn't thought of including the recipe. It's pretty simple, but I think having a dedicated spot for it is a good idea, so I'll go ahead and do that.

src/faq/configuration.md
jacobsky: How did those get there!?!?😱

src/faq/configuration.md
jacobsky: How did those get there!?!?😱 

I'll fix it

src/faq/configuration.md
jacobsky: @Bromeon I meant that the type must inherit from `Script` I'll try to adjust the wording.

jacobsky:
I pushed some changes to fix the issues.

src/recipes/custom-nodes-plugin.md
```diff
@@ -0,0 +1,53 @@
+# Custom Node Plugin
```
Bromeon: Nodes (plural) -- maybe the intro sentence can even be extended with something like "This script allows to register multiple node types, not just one" or so.

src/recipes/custom-nodes-plugin.md
```diff
@@ -0,0 +1,53 @@
...
+    # Add a remove for each registered custom type to clean up
+```
+
+From the editor's main menu, find "Project > Project Settings > Plugins". Find "My Custom Nodes" in the list of plugins and activate this. From here, you can create these directly from the "Create New Node" dialog. In addition, any nodes created this way will disallow the script from being changed or removed.
+
+For more information about this kind of plugin, please refer to the [Godot Documentation](https://docs.godotengine.org/en/stable/tutorials/plugins/editor/making_plugins.html#a-custom-node).
+
+**Note**: This method only works for types that inherit from [`Script`](https://docs.godotengine.org/en/stable/classes/class_script.html). In addition, changing the path to the ".gdns" file will cause the custom `Node` to no longer register correctly.
+
+**Note 2**: This does not register your custom classes in the class database. In order to instantiate the script from GDScript, it will still be necessary to use `var ScriptType = preload("res://path/to/my_node.gdns)` before attempting to instantiate it.
```
Bromeon: ```var ScriptType = preload("res://path/to/my_node.gdns)``` is missing a quote at the end of the path.

jacobsky: Oops. I'll fix that >.<;;

Bromeon:
Thanks again for your continuous efforts to improve the book! 🍾

PR #36, opened 2021-09-20
Game Architecture

Bromeon:


src/gdnative-overview/architecture.md
```diff
@@ -0,0 +1,84 @@
+# Game architecture
+
+_This chapters assumes that you are developing a game with Godot and Rust, however many of the points apply to other projects like simulations or visualizations._
```
jacobsky: As this is two independent clauses you should use a semi-color like the following

... with Godot and Rust; however, many of the points ...

src/gdnative-overview/architecture.md
```diff
@@ -0,0 +1,84 @@
...
+
+For users new to the godot-rust binding, there are a few questions that almost always come up:
+* How do I organize my Rust code the best way?
+* Should I still use GDScript or do everything in Rust?
+* Where should I write my game logic?
+* How can I use the Godot scene tree, if Rust has no inheritance?
+
+With respect to architecture, godot-rust offers a lot of freedom and does not force a certain model upon you. How much you want to develop in GDScript and how much in Rust, is entirely up to you. The choice may depend on your experience, the amount of existing code you already have in either language, the scope of your game or simply personal preference. Each language has their own strengths and weaknesses:
```
jacobsky: This is a great section. Small nitpicks

"With respect to architecture, " might be more readable as "Regarding architecture, godot-rust offers a lot of freedom and does not force you into certain patterns or architectures."


I would also put "Each language has their own strengths and weaknesses:" on a separate line to emphasize that were doing a proper comparison next.

src/gdnative-overview/architecture.md
```diff
@@ -0,0 +1,84 @@
...
+
+* **GDScript** is close to the engine, allows for very fast prototyping and integrates well with the editor. However, its type system is limited and refactoring is often manual. There is no dependency management.
+* **Rust** focuses on type safety, performance and scalability, with mature tooling and ecosystem. The language is rather complex and enforces discipline, and Godot-related tasks tend to be more verbose.
+
+To give you some guide, this chapter highlights three typical patterns. That does not mean you have to adhere to any of them; sometimes hybrid solutions or entirely different designs are worth considering. The three patterns are listed in ascending order w.r.t. complexity and scalability.
```
jacobsky: I like that this area also pointed out everything is intended to highlight some typical patterns while also mentioning that there is room for both.

I made a few tweaks for readability. I think the two most important points were to include that "entirely different designs are also worth considering" so that it feels a little more inclusive and to avoid using acronyms (I wasn't familiar with what "w.r.t." meant initially, so writing out the full word may be more immediately obvious.

A sample for my thoughts is below.

"As a starting point, this chapter highlights three common patterns that have been successfully used with godot-rust. This does not mean you must adhere to any of them; depending upon your system hybrid solutions or entirely different designs are also worth considering. The three patterns are listed with respect to complexity and scalability."

src/gdnative-overview/architecture.md
```diff
@@ -0,0 +1,84 @@
...
+
+
+## 1. Godot game + Rust module
+
+In this architecture, you develop your game primarily in the Godot engine. Most of the game logic resides in GDScript, and the Godot editor is your tool of choice. 
+
+Now, you figure out there is a feature which you don't want to implement in GDScript. Maybe because of performance, or maybe because you want to pull in functionality from the Rust ecosystem, for example libraries for pathfinding, physics, etc.
```
jacobsky: This is a little wordy, so how about the following?

During development, you encounter a feature which you wish to develop in Rust while making this accessible to GDScript. Reasons may include:
- The code is performance critical and GDScript was not fast enough.
- There is a Rust based library that you wish to use, for example, pathfinding, AI, Physics
- You have a segment of your code with high complexity that is difficult to manage in GDScript

src/gdnative-overview/architecture.md
```diff
@@ -0,0 +1,84 @@
...
+
+![Godot scene tree with GDScript, calling to an external Rust module](rust-module.png)
+
+Pros:
+* Very easy to get started, especially for existing Godot codebases.
+* You can fully benefit from Godot's scene graph and the tooling around it.
+* You can test the Rust functionality independently, without running Godot.
+
+Cons:
+* As most of the game logic is in GDScript, you cannot benefit from Rust's features, type safety and refactoring capabilities.
```
jacobsky: It might be more natural to say "and you receive limited benefits from Rust's features"

src/gdnative-overview/architecture.md
```diff
@@ -0,0 +1,84 @@
...
+* Your game logic needs to fit Godot's scene graph model and you have little control about architecture.
+
+
+## 2. Godot scene tree + native scripts
+
+When developing a game in Godot alone, the engine encourages a certain pattern: each game entity is represented by a scene node, and the logic for each node is implemented in a GDScript file. With godot-rust, you can follow the same architecture, except that scripts are implemented in Rust.
```
jacobsky: I feel like the following may be a little more clear.

"Godot Engine encourages a certain pattern: "

"You can follow the same architecture with godot-rust, the only difference is that the scripts are implemented in Rust."

src/gdnative-overview/architecture.md
```diff
@@ -0,0 +1,84 @@
...
+
+
+## 3. Rust game + Godot I/O layer
+
+This architecture is the counterpart to section 1. Most of your game is written in Rust, and you use the engine primarily for input/output handling. You can have an entry point to your Rust library (the _Controller_) which coordinates the simulation.
+
+A typical workflow is as follows:
+1. **Input:** You use the Godot engine to collect user input and events (key pressed, network packet arrived, timer elapsed, ...).
+2. **Processing:** The collected input is passed to Rust. The Controller runs a step in your game simulation and produces results.
+3. **Output:** These results are passed back to Godot and take effect in the scene (node positions, animations, sound effects, ...)
```
jacobsky: needs a fullstop at the end.

src/gdnative-overview/architecture.md
```diff
@@ -0,0 +1,84 @@
...
+
+This might be the most scalable workflow, but it also needs a lot of discipline, and you have to do certain things manually that the other options give you for free.
+
+
+![Game in Rust, Godot scene tree with glue code in GDScript](rust-game.png)
+
+Pros:
+* You are completely free how you organize your Rust game logic. You can have your own entity hierarchy, use an ECS, or simply a few linear collections. 
+* As your game logic runs purely in Rust, you don't need Godot to run the simulation. This allows for the following:
+   * Rust-only headless server
```
jacobsky, line 83: Next 3 are missing the fullstop at the end. That's probably fine so long as it's consistent within the same bulleted list (imo), I just wanted to point it out in case it was a miss.

src/gdnative-overview/architecture.md
```diff
@@ -0,0 +1,84 @@
...
+This architecture is the counterpart to section 1. Most of your game is written in Rust, and you use the engine primarily for input/output handling. You can have an entry point to your Rust library (the _Controller_) which coordinates the simulation.
+
+A typical workflow is as follows:
+1. **Input:** You use the Godot engine to collect user input and events (key pressed, network packet arrived, timer elapsed, ...).
+2. **Processing:** The collected input is passed to Rust. The Controller runs a step in your game simulation and produces results.
+3. **Output:** These results are passed back to Godot and take effect in the scene (node positions, animations, sound effects, ...)
+
+If you follow this pattern strictly, the Godot scene graph can be entirely derived from your Rust state, as such it is only a visualization. There will be some GDScript glue code to tweak graphics/audio output.
+
+This might be the most scalable workflow, but it also needs a lot of discipline, and you have to do certain things manually that the other options give you for free.
```
jacobsky: This workflow can be the most scalable and "Rusty" workflow, but this also requires a lot of discipline. In addition, you will lose certain functionality in the editor, such as the Node tree, without additional work.

src/gdnative-overview/architecture.md
```diff
@@ -0,0 +1,84 @@
...
+It often makes sense to be pragmatic and not try to do everything in Rust. For example, tweaking parameters for particle emitters or animation works much better in GDScript and/or the Godot editor.
+
+![Rust NativeScripts directly inside the Godot tree](in-tree.png)
+
+Pros:
+* You can make full use of Godot's scene graph architecture while still writing the logic in Rust.
+* Existing code and concepts from GDScript can be carried over quite easily.
+
+Cons:
+* You have little architectural freedom and need to orient your entities to Godot's scene tree model.
```
jacobsky: Perhaps something like: "You are generally constrained to Godot's SceneTree model".

src/gdnative-overview/architecture.md
```diff
@@ -0,0 +1,84 @@
...
+* godot-rust is needed throughout your entire code base, tightly coupling your game logic to Godot. This means that testing and isolating functionality will be harder, and you may need to adjust large parts of the code as godot-rust or Godot update versions.
```
jacobsky: This is really good, but it might flow better with something like the following:

"As godot-rust is used throughout your entire codebase, this will tightly couple your game logic to Godot. Testing and isolating functionality can be harder and you are more subject to Godot's version changes."

jacobsky (changes requested):
Thank you for the hard work on this PR. This is a great article and I think the images help tremendously with understanding the whole thing.

Most of my comments are on the nitpicky side, feel free to disregard any that you disagree with substantially, I just wanted to make sure to add all of my thoughts.

Great section!

src/gdnative-overview/architecture.md
Bromeon: Thanks!
I would maybe write:

"Regarding architecture, godot-rust offers a lot of freedom and does not force you into certain patterns ~or architectures~."

As we have the repetition otherwise.

src/gdnative-overview/architecture.md
Bromeon: Great improvement, I'll adopt this!
Regarding "depending upon your system", would it make sense to use "needs" or "requirements" instead of "system"?

And "The three patterns are listed with respect to complexity and scalability."
Does this imply ascending order?

src/gdnative-overview/architecture.md
Bromeon: Good points, I'll replace it! Also I like the 3rd bullet point 🙂

I would do minor changes in wording/punctuation:
* "while making ~this~ **it** accessible to GDScript."
* "Rust-based"
* "for example pathfinding, AI or physics"

Makes sense?

src/gdnative-overview/architecture.md
Bromeon: Is "to receive benefit from" more natural than "to benefit from"?
It sounds to me like literally obtaining benefits (e.g. financial), while "to benefit from" sounds like having an advantage (i.e. what we want to say). But I'm not a native speaker...

src/gdnative-overview/architecture.md
Bromeon: Here it's just keywords, not full sentences, so I think it's OK.
I agree though that we should try to be consistent within one bullet list.

src/gdnative-overview/architecture.md
Bromeon: Here I didn't want to explicitly mention drawbacks, as they are part of the bullet list already, but I wanted to make clear that the other approaches give some things for free which one has to do by hand in this case.

Maybe:

> This can be the most scalable and "Rusty" workflow, but it also requires a lot of discipline. You will need to handcraft some glue code that the other workflows offer for free.

src/gdnative-overview/architecture.md
Bromeon: Great wording! Also added "Godot's **and godot-rust's** version changes" 👍🏼

Bromeon:
Thanks a lot for the feedback! Pushed an amended version as a separate commit.

src/gdnative-overview/architecture.md
jacobsky: That's perfect!

src/gdnative-overview/architecture.md
jacobsky: >  Regarding "depending upon your system", would it make sense to use "needs" or "requirements" instead of "system"?
Yes, "depending upon your needs" or "depending upon your requirements" both flow a lot better.

> And "The three patterns are listed with respect to complexity and scalability."
> Does this imply ascending order?

After thinking about it, while it does imply ordering it doesn't necessary imply which kind of ordering. So doing "The three patterns are listed in ascending order with respect to complexity and scalability" is probably the most clear.

src/gdnative-overview/architecture.md
jacobsky: Sounds perfect :)

src/gdnative-overview/architecture.md
jacobsky: That makes a lot of sense. I think your second version perfectly and succinctly describes that additional work required.

src/gdnative-overview/architecture.md
jacobsky: I think both are alright. I think "you will also be unable to receive many benefits of Rust's features" is also perfectly valid.

On second though, I think that you can probably keep it the way it is as well. I might have been a bit on the pedantic side because you will _technically_ gain some benefits from Rust (scripts written in Rust get those benefits), but the project as a whole will probably not.

I think we should mention the project in this setence.

"As most of the game logic is in GDScript, your project will not benefit from Rust's features, type safety and refactoring capabilities.

jacobsky (changes requested):
Thank you for the second round of changes! I think this is pretty close to perfect. I added some comments on the previously unresolved points.

Bromeon:
Adjusted the discussed point. If the rest is good, I'll squash and merge.

jacobsky (approved):
Looks good to me! Thank you for the hard work :)

PR #40, opened 2021-10-25
Add info about property attribute options

mivort:
This PR expands article about property export with examples for default value setting, creation of no-editor properties and value access/modification hooks.

It might come in handy for people migrating from GDScript.

### DIFF
--- a/src/rust-binding/properties.md
+++ b/src/rust-binding/properties.md
@@ -18,3 +18,71 @@ print("enemies: ", api.enemy_count)
 ```
 
 That's it.
+
+## Export options
+
+The `#[property]` attribute can accept a several options to refine the export behavior.
+
+You can specify default property value with the following argument:
+
+```rust
+#[property(default = 10)]
+enemy_count: i32,
+```
+
+If you need to hide this property in Godot editor, use `no_editor` option:
+
+```rust
+#[property(no_editor)]
+enemy_count: i32,
+```
+
+## Property access hooks
+
+You can add hooks to call methods before and after property value was retrieved or changed.
+
+Note that unlike [GDScript's `setget` keyword](https://docs.godotengine.org/en/3.3/getting_started/scripting/gdscript/gdscript_basics.html?#setters-getters), this does _not_ register a custom setter or getter. Instead, it registers a callback which is invoked _before or after_ the set/get occurs, and lacks both parameter and return value.
+
+```rust
+#[derive(NativeClass, Default)]
+#[inherit(Node)]
+struct GodotApi {
+    // before get
+    #[property(before_get = "Self::before_get")]
+    prop_before_get: i32,
+
+    // before set
+    #[property(before_set = "Self::before_set")]
+    prop_before_set: i32,
+
+    // after get
+    #[property(after_get = "Self::after_get")]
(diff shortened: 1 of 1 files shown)


src/rust-binding/properties.md
```diff
@@ -18,3 +18,81 @@ print("enemies: ", api.enemy_count)
...
+
+```rust
+# #[derive(NativeClass)]
+# #[inherit(Node)]
+# pub struct GodotApi {
+#
+#[property(default = 10)]
+enemy_count: i32,
+#
+# }
```
Bromeon: There are too many `#` line prefixes (this is Markdown, not Rustdoc). Also below.

src/rust-binding/properties.md
```diff
@@ -18,3 +18,81 @@ print("enemies: ", api.enemy_count)
...
+    #[property(after_set = "Self::after_set")]
+    prop_after_set: i32,
+}
+
+impl GodotApi {
+    fn new(_owner: &Node) -> Self {
+        Self::default()
+    }
+
+    fn before_get(&self, owner: TRef<Node>) {
```
Bromeon: Should be `_owner` (also below).

src/rust-binding/properties.md
```diff
@@ -18,3 +18,81 @@ print("enemies: ", api.enemy_count)
...
+#
+#[property(no_editor)]
+enemy_count: i32,
+#
+# }
+```
+
+## Property access hooks
+
+You can add hooks to call methods before and after property value was retrieved or changed.
```
Bromeon, line 42: Maybe add this for clarification:

> Note that unlike [GDScript's `setget` keyword](https://docs.godotengine.org/en/3.3/getting_started/scripting/gdscript/gdscript_basics.html?#setters-getters), this does _not_ register a custom setter or getter. Instead, it registers a callback which is invoked _before or after_ the set/get occurs, and lacks both parameter and return value.

Bromeon (changes requested):
Thank you for the addition!

There's also [some documentation](https://godot-rust.github.io/docs/gdnative/derive/derive.NativeClass.html#property) in the API already, but I don't think it hurts to have larger examples in the book.

Added some comments.

src/rust-binding/properties.md
mivort: That's a very nice note indeed! Added.

src/rust-binding/properties.md
mivort: Done, added `_` prefixes.

src/rust-binding/properties.md
mivort: I've added these as hidden lines to make example 'a bit' more complete, but since it still doesn't compile with the Run button, I suppose there's no need for that... removed it.

Bromeon (approved):
Looks good! If you could [combine the commits into a single one](https://stackoverflow.com/q/5189560), it should be ready to merge 🙂

mivort:
Done, squashed all appends into single commit :)

jacobsky (approved):
Thank you for the additional examples! This is a great addition to the properties section (which I always felt was a bit lacking).

After the latest changes it looks good to me.

PR #43, opened 2021-10-27
Did a second pass at the intro-page of the book.

jacobsky:
Added some additional introduction text.
Added a follow-up question to the FAQ based on https://github.com/godot-rust/godot-rust/issues/798

### DIFF
--- a/src/faq/code.md
+++ b/src/faq/code.md
@@ -277,6 +277,8 @@ Also, you can always put a Mutex<HashMap> into a Rust static and load everything
 
 ## How do I keep a reference of `Node`?
 
+The idiomatic way to maintain a reference to a node in the SceneTree from Rust is to use `Option<Ref<T>>`. 
+
 For example, in the following GDScript code
 ```gdscript
 extends Node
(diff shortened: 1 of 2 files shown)


src/faq/code.md
```diff
@@ -313,6 +315,51 @@ Note: As `TRef<T>` is a temporary pointer, it will be necessary to get the base
...
 This can be done with the [`TRef<T>::claim()`](https://docs.rs/gdnative/latest/gdnative/struct.TRef.html#method.claim) function that will return the persistent version of the pointer that you can store in your class.
 
+## What is the Rust equivalent to `onready var` in GDScript
+
+Rust does not have a direct equivalent to `onready var`. The most idiomatic workaround with Rust is to use `Option<Ref<T>>` of you need the Godot node type or `Option<Instance<T>>` if you are using a Rust based `NativeClass`.
+
+```gdscript
+extends Node
+class_name MyClass
+onready var node = $node2d
```
Bromeon: This seems to be syntax-highlighted weirdly by GitHub. Does `$Node2D` improve it?
I think the latter is also the default name when created in the Godot editor.

src/introduction.md
```diff
@@ -4,8 +4,27 @@ Welcome to the `godot-rust` book! This is a work-in-progress user guide for the
...
 If you have used earlier versions of `godot-rust` before, see [Migrating from godot-rust 0.8.x](./migrating-0-8.md) for a quick guided tour to the new API.
 
+## About godot-rust
+
+This project specifically supports the [Rust Programming Language](https://www.rust-lang.org/) bindings to the [GDNative API](https://docs.godotengine.org/en/stable/tutorials/plugins/gdnative/gdnative-cpp-example.html) for the Godot Game Engine.
+
+Outside of personal preference, Rust may be a good choice for your game for some of the following reasons:
+- Native levels of performance.
+- Memory safety validated at compile time.
+- [Fearless Concurrency](https://blog.rust-lang.org/2015/04/10/Fearless-Concurrency.html).
```
Bromeon: Although it might be fair to say that those two points are somewhat compromised when having some code in GDScript, or using the GDNative API (at least in its current form).

Do you think we should mention it? If so, I would definitely not add more than a side note (e.g. before the sentence `For more information about...`).

Bromeon (commented):
Looks good, I only have a question about the thread/memory safety (see comment).
Thanks a lot!

src/faq/code.md
jacobsky: Ah, that's right, I'm the odd one out since I use snake_case for everything. I'll fix that.

src/introduction.md
jacobsky: That's a good point, those points are only valid when using Rust exclusively, so I will amend that.

jacobsky:
@Bromeon I have made the fixes that you recommended. If there's no objections to the new text, I'd be happy to merge this in.

Bromeon:
Thank you very much! 🙂

PR #44, opened 2021-10-27
Tokio runtime recipe

lyonbeckers:
Here's a little recipe describing how I use a tokio runtime with the `gdnative-async` changes

### DIFF
--- a/src/recipes/tokio_runtime.md
+++ b/src/recipes/tokio_runtime.md
@@ -1,0 +1,97 @@
+# Tokio Runtime
+
+This recipe is based off of the test written for `gdnative-async`, which uses the `futures` crate in the executor. For cases where you may need a `tokio` runtime, it is possible to execute spawned tokio tasks in much the same way, with some alterations.
+
+## Defining the Executor
+The executor itself can be defined the same way.
+
+```rust
+thread_local! {
+	static EXECUTOR: &'static SharedLocalPool = {
+		Box::leak(Box::new(SharedLocalPool::default()))
+	};
+}
+```
+
+However, our `SharedLocalPool` will store a `LocalSet` instead, and the `futures::task::LocalSpawn` implementation for the type will simply spawn a local task from that.
+
+```rust
+use tokio::task::LocalSet;
+
+#[derive(Default)]
+struct SharedLocalPool {
+	local_set: LocalSet,
+}
+
+impl futures::task::LocalSpawn for SharedLocalPool {
+	fn spawn_local_obj(
+		&self,
+		future: futures::task::LocalFutureObj<'static, ()>,
+	) -> Result<(), futures::task::SpawnError> {
+		self.local_set.spawn_local(future);
+
+		Ok(())
+	}
+}
+```
+
+## The Executor Driver
+
+Finally, we need to create a `NativeClass` which will act as the driver for our executor. This will store the tokio `Runtime`.
+
+```rust
+use tokio::runtime::{Builder, Runtime};
+
+#[derive(NativeClass)]
+#[inherit(Node)]
+struct AsyncExecutorDriver {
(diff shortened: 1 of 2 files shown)


src/recipes/tokio_runtime.md
```diff
@@ -0,0 +1,98 @@
...
+#[methods]
+impl AsyncExecutorDriver {
+	#[export]
+	fn _process(&self, _owner: &Node, _delta: f64) {
+		EXECUTOR.with(|e| {
+			self.runtime
+				.block_on(async {
+					e.local_set
+						.run_until(async {
+							tokio::task::spawn_local(async {}).await
```
jacobsky, line 84: Some additional clarification for folks not familiar with the runtime, in this case the task code belongs inside the closure defined by the `tokio::task::spawn_local`'s async block, correct?

i.e.

```rs
 tokio::task::spawn_local( async { /* my custom code */})
 ```
 
 Would it be possible to get add a comment for clarity?

File: src/recipes/tokio_runtime.md [hunk 2]
```diff
@@ -0,0 +1,97 @@
+# Tokio Runtime
+
+This recipe is based off of the test written for `gdnative-async`, which uses the `futures` crate in the executor. For cases where you may need a `tokio` runtime, it is possible to execute spawned tokio tasks in much the same way, with some alterations.
+
```
jacobsky, line 4: Out of curiousity, what version of tokio is this example using?

I generally like to add which versions of any external framework the recipes are using. In this case it may also be a good idea to mention that compiling `gdnative` with `async` requires working off of the gdnative master branch for github and that the "async" feature flag is required.

Something like the following:

> ## Requirements
> This recipe requires the following entries in your Cargo.toml file
> 
> ```toml
> tokio = "x.y.z"
> gdnative = { git = "https://github.com/godot-rust/godot-rust.git", features = ["async"]}
> ```

If you'd like to use the above as is, you can just copy-paste the markdown text using the quote reply and copying it from there :)

jacobsky (changes requested):
Thank you for the PR! This is an amazing recipe and I'm sure that this will be the perfect place to point folks that are curious about dipping their toes into using async with godot-rust.

src/recipes/tokio_runtime.md
lyonbeckers: This is where things get a little confusing actually, but I think I can write a bit more to clarify it. `run_until` will automatically run the tasks that have been spawned in the `local_set`. To the best of my understanding, based off of the tokio docs, we then need to spawn a new local task to ensure that the future is spawned on the local task set. Not sure if I'm understanding 100% correctly, but without it, the other spawned tasks don't run.

~~However, now that I'm looking at the docs again, there may even be a simpler way to do this, so I'll try that first~~
EDIT: never mind, what I was thinking of would await all tasks to completion

src/recipes/tokio_runtime.md
lyonbeckers: Very good point, especially since we require a feature for tokio as well. Added

src/recipes/tokio_runtime.md
lyonbeckers: Added a bit more clarification, hopefully this makes sense

jacobsky (approved):

jacobsky:
Thank you for the clarifications! This looks good to me!

@Bromeon any nitpicks from your end or should I just go ahead and merge this?

Bromeon:
Awesome, thanks a lot for taking the time to write it down!

Minor formatting things (e.g. tabs -> spaces), but no dealbreaker. There are anyway some formatting inconsistencies I wanted to address in other pages of the book, so I can as well do that in a separate PR, should be more efficient.

For me, we can merge it! 🚀

PR #52, opened 2021-10-30
Few minor fixes

bluenote10:
I haven't used godot-rust in a long time, and having a look into the book again, it is great to see how it has evolved!

Just fixing a few tiny things I noticed while re-reading...

### DIFF
--- a/src/exporting/macosx.md
+++ b/src/exporting/macosx.md
@@ -137,5 +137,5 @@ Double-check your `.dylib` file is there.
 
 ## Useful links
 
-* https://github.com/tpoechtrager/osxcross : tool used to install the Mac OS X SDK on Linux
-* https://wapl.es/rust/2019/02/17/rust-cross-compile-linux-to-macos.html : a complete tutorial on how to use osxcross
+* [https://github.com/tpoechtrager/osxcross](https://github.com/tpoechtrager/osxcross): tool used to install the Mac OS X SDK on Linux
+* [https://wapl.es/rust/2019/02/17/rust-cross-compile-linux-to-macos.html](https://wapl.es/rust/2019/02/17/rust-cross-compile-linux-to-macos.html): a complete tutorial on how to use osxcross
--- a/src/faq/code.md
+++ b/src/faq/code.md
@@ -40,7 +40,7 @@ impl SignalEmitter {
 }
 ```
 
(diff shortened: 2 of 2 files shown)


src/exporting/macosx.md
```diff
@@ -137,5 +137,5 @@ Double-check your `.dylib` file is there.
 
 ## Useful links
 
-* https://github.com/tpoechtrager/osxcross : tool used to install the Mac OS X SDK on Linux
-* https://wapl.es/rust/2019/02/17/rust-cross-compile-linux-to-macos.html : a complete tutorial on how to use osxcross
+* [https://github.com/tpoechtrager/osxcross](https://github.com/tpoechtrager/osxcross): tool used to install the Mac OS X SDK on Linux
```
jacobsky, line 140: Thanks for the catch. I'd never actually tried to click those "links" before, so I never saw that.

bluenote10: Often there is also a switch in the markdown to html conversion to automatically insert hyperlinks for plain URLs, but it is maybe easier to handle it manually in this case.

src/faq/code.md
```diff
@@ -85,19 +85,23 @@ Here is an example of some common `unsafe` usage that you will often see and use
 fn get_a_node(&self, owner: TRef<Node>) {
     // This is safe because it returns an option that Rust knows how to check.
     let child = owner.get_child("foo");
```
jacobsky, line 87: I like this series of tweaks. Regarding the line breaks, was there a specific tool that you ran to get the linebreaks? Or was it just a manual endeavor to keep the comment lines to a certain length?

jacobsky (approved):
Thank you for the PR (Sorry for the fragmented review). I just have the two comments related to this and I'll go ahead and merge it in. :)

src/faq/code.md
bluenote10: Unfortunately just an old-school manual endeavor :wink: I broke it around the width that is visible for me:

image

jacobsky: Thank you for the input! I checked it out on my end and it's definitely more readable. Thank you for the help!

jacobsky:
@Bromeon If there are no objections, I'll go ahead and merge this PR.

Bromeon (approved):
Good with me!

jacobsky:
Thank you again for the PR!

Issue #54, opened 2021-11-01
[FAQ Entry] How to implement polymorphic functions?


bluenote10:
This would be a possible candidate as an FAQ entry: _How to implement polymorphic functions?_

For instance, the goal is to implement a free-floating helper function that can operate on any `Control` like:

```rust
pub fn set_anchor_full_rect_non_polymorphic(control: &Ref<Control, Unique>) {
    control.set_anchor(GlobalConstants::MARGIN_RIGHT, 0.0, false, true);
    control.set_anchor(GlobalConstants::MARGIN_BOTTOM, 0.0, false, true);
    control.set_anchor(GlobalConstants::MARGIN_RIGHT, 1.0, false, true);
    control.set_anchor(GlobalConstants::MARGIN_BOTTOM, 1.0, false, true);
}
```

Due to the explicit use of `Control` this isn't polymorphic though.

I assume this can be solved using `SubClass`. My naive attempt

```rust
pub fn set_anchor_full_rect<T>(control: &Ref<T, Unique>)
where
    T: GodotObject<RefKind = ManuallyManaged> + SubClass<Control>,
{
    control.set_anchor(GlobalConstants::MARGIN_RIGHT, 0.0, false, true);
    control.set_anchor(GlobalConstants::MARGIN_BOTTOM, 0.0, false, true);
    control.set_anchor(GlobalConstants::MARGIN_RIGHT, 1.0, false, true);
    control.set_anchor(GlobalConstants::MARGIN_BOTTOM, 1.0, false, true);
}
```

doesn't quite work though (no method named `set_anchor` found for reference `&godot::Ref<T, godot::prelude::Unique>`), and probably using `Unique` isn't ideal in terms of re-usability.

jacobsky:
@bluenote10 Thanks for the idea. I think this would be a good candidate for the book as it's a solid usecase.

Just to clarify, the question is essentially "How do to access to the subclass functions?", correct?

For example: From a `Ref<Control>` how do you access a the `CanvasItem` functions?

In the example above, you should be able to use the subclass, but you would need to add a `let control = control.cast::<Control>` to explicitly [cast the type](https://docs.rs/gdnative/0.9.3/gdnative/struct.TRef.html#method.cast) into the type that uses it.

bluenote10:
> Just to clarify, the question is essentially "How do to access to the subclass functions?", correct?

What I meant is: How can I pass in any class that is a subclass of `Container`?

[...]

PR #55, opened 2021-11-03
Added faq entry on runtime polymorphism

bluenote10:
I'll give #54 a shot, feel free to adjust as needed ;)

I wasn't sure if this FAQ entry is the right place to explain `Deref` with its limitation, so I kept it at the bare minimum for now.

### DIFF
--- a/src/faq/code.md
+++ b/src/faq/code.md
@@ -325,6 +325,34 @@ Note: As `TRef<T>` is a temporary pointer, it will be necessary to get the base 
 This can be done with the [`TRef<T>::claim()`](https://docs.rs/gdnative/latest/gdnative/struct.TRef.html#method.claim) function that will return the persistent version of the pointer that you can store in your class.
 
 
+## How to implement function arguments that accept any Godot subclass (runtime polymorphism)?
+
+Let's assume you want to implement a helper function that should accept any kind of [`Container`](https://docs.godotengine.org/en/stable/classes/class_control.html#class-control). This can be solved by a combination of `SubClass` and `upcast`:
+
+```rust
+fn do_something_with_container<T>(container: TRef<'_, T>)
+	where T: GodotObject + SubClass<Container>
+{
+    // First upcast to a true container:
+	let container = container.upcast();
+    // Now you can call `Container` specific methods like:
+    container.set_size(...);
+}
+```
+
+This function can now be used with arbitrary subclasses, for instance:
+
+```rust
+fn some_usage() {
+	let panel_container: Ref<PanelContainer> = PanelContainer::new().into_shared();
+	let panel_container: TRef<PanelContainer> = unsafe { panel_container.assume_safe() };
+	do_something_with_container(panel_container);
+}
+```
+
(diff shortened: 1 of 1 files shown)


src/faq/code.md
```diff
@@ -325,6 +325,34 @@ Note: As `TRef<T>` is a temporary pointer, it will be necessary to get the base
 This can be done with the [`TRef<T>::claim()`](https://docs.rs/gdnative/latest/gdnative/struct.TRef.html#method.claim) function that will return the persistent version of the pointer that you can store in your class.
 
 
+## How to implement function arguments that accept any Godot subclass (runtime polymorphism)?
```
jacobsky: In this case it may be more consistent to change the question to "How do I..."? as that is consistent with other questions like "How do I keep a reference of Node?", etc.

src/faq/code.md
```diff
@@ -325,6 +325,34 @@ Note: As `TRef<T>` is a temporary pointer, it will be necessary to get the base
...
+
+Let's assume you want to implement a helper function that should accept any kind of [`Container`](https://docs.godotengine.org/en/stable/classes/class_control.html#class-control). This can be solved by a combination of `SubClass` and `upcast`:
```
jacobsky: This paragraph can probably be more clear if we split it into two sections:
1. The more generic answer like: "Access to the `SubClass` of a `TRef<T>` may be accessed via the `upcast` method.
2. An example:; "For example: if you wanted to implement a helper function..."

jacobsky (changes requested):
Thank you for the PR! I greatly appreciate you bringing our attention to this subject in #54 

Overall this entry looks good, I just have two minor nitpicks related to making the phrasing a little more consistent with the rest of the FAQ.

src/faq/code.md
```diff
@@ -325,6 +325,34 @@ Note: As `TRef<T>` is a temporary pointer, it will be necessary to get the base
...
+
+```rust
+fn do_something_with_container<T>(container: TRef<'_, T>)
+	where T: GodotObject + SubClass<Container>
+{
+    // First upcast to a true container:
+	let container = container.upcast();
+    // Now you can call `Container` specific methods like:
+    container.set_size(...);
+}
```
Bromeon: Could you run the code through rustfmt?
Should have 4 space indentation.

Same for the other snippet.

src/faq/code.md
```diff
@@ -325,6 +325,34 @@ Note: As `TRef<T>` is a temporary pointer, it will be necessary to get the base
...
+## How to implement function arguments that accept any Godot subclass (runtime polymorphism)?
+
+Let's assume you want to implement a helper function that should accept any kind of [`Container`](https://docs.godotengine.org/en/stable/classes/class_control.html#class-control). This can be solved by a combination of `SubClass` and `upcast`:
+
+```rust
+fn do_something_with_container<T>(container: TRef<'_, T>)
+	where T: GodotObject + SubClass<Container>
+{
+    // First upcast to a true container:
+	let container = container.upcast();
```
Bromeon: Maybe annotate the type here explicitly?
That helps people understand what exactly `upcast()` does.

Bromeon (commented):
Thanks a lot for the PR! Added small comments.

src/faq/code.md
bluenote10: I'm a formatting pedantic, how could that have happened to me :joy: (of course: a space vs tabs thing...)

bluenote10:
Thanks for the suggestions, fully agreeing to everything, I hope I've fixed it all now.

jacobsky (approved):
Looks good to me! If you can go ahead and squash the two commits, I'll merge it in. I can also squash it, but then you'd lose the credit for the commits.

bluenote10:
Feel free to use the auto-squash, I don't care about the credits ;)

PR #59, opened 2021-12-20
Custom Godot builds

Bromeon:
Updates the page to reflect new approach in https://github.com/godot-rust/godot-rust/pull/833.


jacobsky (approved):
Thank you for the guide, it appears simple enough! I think that this should be a good starting point. I haven't had too much time to test the custom builds for godot, but I will try to do some of that over the new years holidays.

The guide looks great! Thank you for the hard work!

PR #62, opened 2022-02-12
FAQ: bindgen C headers not found

Bromeon:
Moved from godot-rust ReadMe, in https://github.com/godot-rust/godot-rust/pull/856.


jacobsky:
Thank you for this! Everything looks good! It also renders nicely on my PC too :)

PR #63, opened 2022-03-12
Fixes incorrect getting-started docs

insou22:


Bromeon:
To be fair, the entire section is outdated. The template project has been moved to [macalimlim/godot-rust-template](https://github.com/macalimlim/godot-rust-template), i.e. building the project should be explained there and not in the book.

Could you instead replace the "Using the template" section with something like 
> One way to get started with godot-rust is a full-fledged (inofficial) template, which can be found [here](https://github.com/macalimlim/godot-rust-template).

Bromeon:
Closed in favor of #65

PR #65, opened 2022-03-15
Updated setup.md

ShayanGsh:
Basically #63, but also added cargo-make in the setup section, and did what @Bromeon had suggested on #63. I hope this'd be useful.

### DIFF
--- a/src/getting-started/setup.md
+++ b/src/getting-started/setup.md
@@ -47,9 +47,9 @@ clang -v
 
 ## Using the template
 
-We also provide an experimental [template](https://github.com/godot-rust/godot-rust-template) to get you started right away. All the boilerplate stuff is already done for you. If you have decided to use the template, you can skip section 2.2. Check out the [wiki](https://github.com/godot-rust/godot-rust-template/wiki) for instructions.
+One way to get started with godot-rust is a full-fledged (inofficial) template, which can be found [here](https://github.com/macalimlim/godot-rust-template) to get you started right away. All the boilerplate stuff is already done for you. If you have decided to use the template, you can skip section 2.2. Check out the [wiki](https://github.com/godot-rust/godot-rust-template/wiki) for instructions.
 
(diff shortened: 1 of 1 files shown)


Bromeon:
Thank you for the update! As a tip, you can also reuse the same pull request by force-pushing to the branch (`git push --force-with-lease`). Titles and descriptions can also be edited after you've created the PR 🙂 

Some feedback:
* Maybe mention that while a lot of stuff is already done in the template, it requires some extra dependencies and toolchains to be set up. No need to go into details, but just make sure the user knows what to expect.
* I would remove the sections _Setup_ and _Usage_ entirely. The user will need to check out the template's documentation already, and we just risk becoming outdated and needing to chase external changes in the book.

ShayanGsh:
Thank you for the tip, I'm fairly new to PRs as you can probably tell lol

I removed the setup and usage sections, replaced the wiki with the template's wiki, mentioned extra dependencies, also replaced "work-in-progress" with "not maintained by us", it made more sense to me.

src/getting-started/setup.md
```diff
@@ -47,18 +47,8 @@ clang -v
 
 ## Using the template
 
-We also provide an experimental [template](https://github.com/godot-rust/godot-rust-template) to get you started right away. All the boilerplate stuff is already done for you. If you have decided to use the template, you can skip section 2.2. Check out the [wiki](https://github.com/godot-rust/godot-rust-template/wiki) for instructions.
+> If you have decided to use the template, you can skip section 2.2.
```
Bromeon: I think we can remove this line. It applies to virtually any sections, and numeric section references are at the risk of becoming outdated.

Bromeon (commented):
One minor comment, otherwise looks good to me! I like the additional changes 🙂

src/getting-started/setup.md
ShayanGsh: That is true. Even when using the template, going through section 2.2 would be helpful to most people to get started and understand the basics.

Bromeon:
Awesome, thanks a lot!
Your changes are now live: https://godot-rust.github.io/book/getting-started/setup.html#using-the-template

PR #67, opened 2022-05-08
Changes the default build version to 0.10 on the hello-world page

Lambdanaut:
This caused me a bit of a headache when looking at the godot-examples, because they use signals a different way than are used in the older 0.9 version given on this getting started page. 

This change reflects the latest version of godot-rust for noobies so that the godot examples will work for them.


Bromeon:
Very nice, thanks!

There are also quite a few occurrences of docs.rs still referring to `0.9*`. Are you interested in updating those as well?
If yes, would be good to use `latest` instead of just `0.10.0`, otherwise we'll face the same issue next minor version.

grafik

Lambdanaut:
No problem. Done 👍

src/faq/code.md
```diff
@@ -205,7 +205,7 @@ impl StaticUtil {
 }
 ```
 
-[`Aether`](https://docs.rs/gdnative/0.9/gdnative/prelude/struct.Aether.html) is a special user-data wrapper intended for zero-sized types, that does not perform any allocation or synchronization at runtime.
+[`Aether`](https://docs.rs/gdnative/latest/gdnative/prelude/struct.Aether.html) is a special user-data wrapper intended for zero-sized types, that does not perform any allocation or synchronization at runtime.
```
Bromeon: This one is no longer in the prelude, the link would be:
https://docs.rs/gdnative/latest/gdnative/export/user_data/struct.Aether.html

src/rust-binding/to-variant-from-variant-export.md
```diff
@@ -31,7 +31,7 @@ For more information about how you can customize the behavior of the dervive mac
 
 ## Export Trait
 
-The Godot editor retrieves property information from [Object::get_property_list](https://docs.godotengine.org/en/stable/classes/class_object.html#id2). To populate this data, `godot-rust` requires that the [`Export`](https://docs.rs/gdnative/0.9.3/gdnative/nativescript/trait.Export.html) trait be implemented for each type Rust struct.
+The Godot editor retrieves property information from [Object::get_property_list](https://docs.godotengine.org/en/stable/classes/class_object.html#id2). To populate this data, `godot-rust` requires that the [`Export`](https://docs.rs/gdnative/latest/gdnative/nativescript/trait.Export.html) trait be implemented for each type Rust struct.
```
Bromeon, line 34: Also here, new path:
https://docs.rs/gdnative/latest/gdnative/export/trait.Export.html

src/rust-binding/to-variant-from-variant-export.md
```diff
@@ -47,7 +47,7 @@ struct Foo {
 }
 
 impl Export for Foo {
-    // This type should normally be one of the types defined in [gdnative::nativescript::hint](https://docs.rs/gdnative/0.9.3/gdnative/nativescript/init/property/hint/index.html).
+    // This type should normally be one of the types defined in [gdnative::nativescript::hint](https://docs.rs/gdnative/latest/gdnative/nativescript/init/property/hint/index.html).
```
Bromeon:
```
[gdnative::export::hint](https://docs.rs/gdnative/latest/gdnative/export/hint/index.html)
```

Bromeon (commented):
A few path paths have changed in 0.10; I wrote the new links so you don't need to search for them.

Lambdanaut:
Links fixed 👍

src/rust-binding/to-variant-from-variant-export.md
```diff
@@ -31,7 +31,7 @@ For more information about how you can customize the behavior of the dervive mac
 
 ## Export Trait
 
-The Godot editor retrieves property information from [Object::get_property_list](https://docs.godotengine.org/en/stable/classes/class_object.html#id2). To populate this data, `godot-rust` requires that the [`Export`](https://docs.rs/gdnative/0.9.3/gdnative/nativescript/trait.Export.html) trait be implemented for each type Rust struct.
+The Godot editor retrieves property information from [Object::get_property_list](https://docs.rs/gdnative/latest/gdnative/export/trait.Export.html). To populate this data, `godot-rust` requires that the [`Export`](https://docs.rs/gdnative/latest/gdnative/nativescript/trait.Export.html) trait be implemented for each type Rust struct.
```
Bromeon: The `Object::get_property_list` link is now pointing to the wrong thing.

You may want to fix the 2nd link though, which points to `nativescript`.

Lambdanaut: where would we be without your critical eye at the whims of my edits 😩

This link is now correct for the second one as well? https://github.com/godot-rust/book/pull/67/files#diff-869226e3b08a92b24cdfd18ad992b82e54b50bc6decb064d0cd2b209e90840c5R50

(I think the link changed from nativescript in this latest version)

Thanks for your patience!

Bromeon:
Thanks a lot for the continuous fixes! 🙂

PR #68, opened 2022-05-15
Added note about the export attribute rename

KennethWilke:
I was tinkering with some recently updated examples and was a bit
confused by my lack of the `godot` NativeClass attribute, so I thought
it may be a good call our here for others as well.


Bromeon:
Great idea, thanks!

PR #69, opened 2022-05-18
WIP: GDScript built-in alternatives

KennethWilke:
This is my first step towards addressing https://github.com/godot-rust/godot-rust/issues/893

These changes add a new FAQ page that will hopefully be a good quick reference, especially in cases of porting GDScript code to godot-rust. I wanted to narrow the list down of methods to consider for replication/re-implementation, and a wide swath of the math related functions are already provided in Rust's `std` crate.

Would appreciate any thoughts/feedback on this format before fleshing it out a bit more

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -21,6 +21,7 @@
     - [Versioning and supported platforms](./faq/support.md)
     - [Community](./faq/community.md)
     - [Godot 4.0 Status](./faq/godot4.md)
+    - [GDScript method alternatives](./faq/gdscript-builtins.md)
 - [(TODO) Testing](./testing.md)
     - [Structuring code for testing](./testing/structure.md)
     - [Testing with the engine](./testing/engine.md)
--- a/src/faq/gdscript-builtins.md
+++ b/src/faq/gdscript-builtins.md
@@ -1,0 +1,167 @@
+# GDScript method alternatives
+
+Godot's GDScript has several [built-in methods](https://docs.godotengine.org/en/stable/classes/class_@gdscript.html) that provide a range of functionality useful for game design and implementation. For Godot 3.x's [GDNative interface](https://github.com/godotengine/godot-headers), these methods are not exposed, but fret not! Most of the math related functions (and many other useful ones) are part of Rust's `std` crate. 
+
+Many of the other commonly needed built in methods have analogs provided by godot-rust, such as [`godot_print!`](https://godot-rust.github.io/docs/gdnative/log/macro.godot_print.html) and [`GodotObject::from_instance_id()`](https://godot-rust.github.io/docs/gdnative/object/trait.GodotObject.html#method.from_instance_id).
+
+This page hopes to provide a quick reference for finding equivalent functionality for these various methods, which can be particularly useful for porting GDScript code to Rust.
+
+## Table of contents
+<!-- toc -->
+
+## Maths
+
+Since GDScript's [`float` built-in type](https://docs.godotengine.org/en/stable/classes/class_float.html#class-float) is a double-precision floating point, the references below refer to Rust's the 64 bit versions in `std` ([f64 primitive](https://doc.rust-lang.org/stable/std/primitive.f64.html) and [std::f64](https://doc.rust-lang.org/stable/std/f64/index.html)), 32-bit versions of all of these are also provided by [f32](https://doc.rust-lang.org/stable/std/primitive.f32.html) / [std::f32](https://doc.rust-lang.org/stable/std/f32/index.html).
+
+Though the `float` type in GDScript is a double-precision floating-point, single-precision floating-points are also used as mentioned in the [Godot Documentation](https://docs.godotengine.org/en/stable/classes/class_float.html#class-float):
+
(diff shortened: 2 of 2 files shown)


Bromeon:
Thanks a lot for this great effort!

I think this would fit better into the Rust code documentation than the book, as it's a reference to look up concrete symbols, and not really prose or FAQ style. This makes also linking between symbols much easier.

We should probably dedicate a new module for these methods, e.g. `gdnative::global` (name to be determined).
The module-level docs can then mention the details as outlined here. Since it's also Markdown, you could probably just copy-paste most of it, and prefix it with `///` 🙂

What do you think?

KennethWilke:
Hmm, I'm very open to input on where this should go. The book felt like a good starting place, as I'm not sure how/where this would best be formatted into the Rust code docs.

I hadn't pictured it going in the code docs because I was thinking "it's documenting something that's not in there!", but if I follow your train of thought it'd be in the module level documentation rather than associated with some specific function and I think that makes a lot of sense and the symbol linking also does sound like it'd be good to have.

I'll probably keep this PR open as reference until we find its new home. My main goal for this first pass was to considerably shorten the list of methods by taking care of the low hanging fruit so we more simply identify what may be "missing". These are the elements that don't yet have some form of alternative in this bit of markdown, some of which are certainly already in the project:

### GDScript built-in math methods

* cartesian2polar ( float x, float y ) -> Vector2
* db2linear ( float db ) -> float
* decimals ( float step ) -> int
* dectime ( float value, float amount, float step ) -> float
* ease ( float s, float curve ) -> float
* fmod ( float a, float b ) -> float
* fposmod ( float a, float b ) -> float
* inverse_lerp ( float from, float to, float weight ) -> float
* is_equal_approx ( float a, float b ) -> bool
* is_zero_approx ( float s ) -> bool
* lerp ( Variant from, Variant to, float weight ) -> Variant
* lerp_angle ( float from, float to, float weight ) -> float
* linear2db ( float nrg ) -> float
* move_toward ( float from, float to, float delta ) -> float
* nearest_po2 ( int value ) -> int
* polar2cartesian ( float r, float th ) -> Vector2
* posmod ( int a, int b ) -> int
* pow ( float base, float exp ) -> float
* range_lerp ( float value, float istart, float istop, float ostart, float ostop ) -> float
* sign ( float s ) -> float
* smoothstep ( float from, float to, float s ) -> float
* step_decimals ( float step ) -> int
* stepify ( float s, float step ) -> float
* wrapf ( float value, float min, float max ) -> float
* wrapi ( int value, int min, int max ) -> int

## Data/Type Conversions

* Color8 ( int r8, int g8, int b8, int a8=255 ) -> Color
* ColorN ( String name, float alpha=1.0 ) -> Color
* bytes2var ( PoolByteArray bytes, bool allow_objects=false ) -> Variant
* char ( int code ) -> String
* convert ( Variant what, int type ) -> Variant
* dict2inst ( Dictionary dict ) -> Object
* inst2dict ( Object inst ) -> Dictionary
* ord ( String char ) -> int
* parse_json ( String json ) -> Variant
* str ( ... ) vararg -> String
* str2var ( String string ) -> Variant
* to_json ( Variant var ) -> String
* var2bytes ( Variant var, bool full_objects=false ) -> PoolByteArray
* var2str ( Variant var ) -> String

## Random Number Generation

* rand_range ( float from, float to ) -> float
* rand_seed ( int seed ) -> Array
* randf ( ) -> float
* randi ( ) -> float
* randomize ( ) -> void
* seed ( int seed ) -> void

## Debugging / Utility / Other

* assert ( bool condition, String message="" ) -> void
* funcref ( Object instance, String funcname ) -> FuncRef
* get_stack ( ) -> Array
* hash ( Variant var ) -> int
* instance_from_id ( int instance_id ) -> Object
* is_instance_valid ( Object instance ) -> bool
* len ( Variant var ) -> int
* load ( String path ) -> Resource
* print_stack ( ) -> void
* printerr ( ... ) vararg -> void
* printraw ( ... ) vararg -> void
* prints ( ... ) vararg -> void
* printt ( ... ) vararg -> void
* push_error ( String message ) -> void
* push_warning ( String message ) -> void
* range ( ... ) vararg -> Array
* type_exists ( String type ) -> bool
* typeof ( Variant what ) -> int
* validate_json ( String json ) -> String
* weakref ( Object obj ) -> WeakRef
* yield ( Object object=null, String signal="" ) -> GDScriptFunctionState

Bromeon:
Thanks for this categorization!

I agree that this can be left open as a discussion item, until we have a more concrete idea of how to provide `@GDScript` methods in godot-rust.

Things to note:
* There is already a [`GDScript`](https://godot-rust.github.io/docs/gdnative/api/struct.GDScript.html) struct with a different meaning in `gdnative::api`, so naming the module `gdscript` could be slightly confusing, and `api::gdscript` could cause name collisions in the future.
* A similar [`GlobalConstants`](https://godot-rust.github.io/docs/gdnative/api/struct.GlobalConstants.html) exists, which implements the [`@GlobalScope`](https://docs.godotengine.org/en/stable/classes/class_@globalscope.html) Godot API. 

Since these functions are manually written and not generated, they should probably not go into module `gdnative::api`. It could be a new top-level `gdnative::global`. I'm open for alternative naming suggestions.

Bromeon:
Should we maybe discuss the concrete API over at https://github.com/godot-rust/godot-rust/issues/893?

Then we would leave this here as a reference, but it's not great to have the same discussion in two places 😉 
I'll copy my answer over.

Bromeon:
Since https://github.com/godot-rust/godot-rust/pull/901 has been merged, I think we can close this.

Your exhaustive list of methods is still very valuable, but as mentioned, it might make more sense in the code documentation (under `globalscope` module) than in the book. I'll leave it up to you if you have time and motivation to migrate it over to a PR in godot-rust. I think we already have a good base with the existing `globalscope` module anyway 🙂

PR #70, opened 2022-06-16
Fixed '_owner' parameter in FAQ/EntityFactory

tuto193:
The `_owner` parameter changed from `Object` to `Reference`, so it compiles properly.


Bromeon (approved):
Looks good, thanks for the fix!

PR #71, opened 2022-06-25
Add HTML5 export guide

derivator:
@Bromeon as discussed in [#647](https://github.com/godot-rust/godot-rust/issues/647) this should cover most of the current problems with HTML5 exports.

### DIFF
--- a/src/exporting/html5.md
+++ b/src/exporting/html5.md
@@ -5,8 +5,8 @@ Exporting to HTML5 works just like exporting to other platforms, however there a
 ## What you need (to know)
 
 The Godot HTML5 export templates are built with [Emscripten](https://emscripten.org/), so you need to build your Rust code for the `wasm32-unknown-emscripten` target (`wasm32-unknown-unknown` will not work).  
-You need to build your code with a version of Emscripten that is compatible with your version of Rust and (ideally identical to) the version of Emscripten used to build the export template.  
-In practice, this means you might have to:
+Since Emscripten does not offer a stable ABI between versions, your code and the Godot export template have to be built with the same Emscripten version, which also needs to be compatible with the Rust compiler version you are using.
+In practice, this means you probably have to do some or all of these things:
 * install a specific version of Emscripten
 * [build the HTML5 export template yourself](https://docs.godotengine.org/en/stable/development/compiling/compiling_for_web.html) 
 * use a specific version of Rust
@@ -18,20 +18,23 @@ emcc -v
 rustc --version --verbose
 ```
 
-A version mismatch between Rust and Emscripten will probably result in compiler errors, while incompatible Emscripten versions between Rust and Godot will result in cryptic runtime errors such as `indirect call signature mismatch`.
+A version mismatch between Rust and Emscripten will probably result in compiler errors, while using incompatible Emscripten versions between Rust and Godot will result in cryptic runtime errors.
+
+Emscripten uses a cache to store build artifacts and header files. In some cases this means that the order in which steps are completed matters.
 
 Also note that `wasm32-unknown-emscripten` is a 32-bit target, which can cause problems with crates incorrectly assuming that `usize` is 64 bits wide.
 
-## Godot 3.4.4
+## Godot 3.5
 
 **Disclaimer**: _Currently, the following steps are only tested and confirmed to work on Linux._
 
-Godot 3.4.4's export template is built with a very old Emscripten version that doesn't play nice with Rust. You will have to build it yourself with a more recent version.  
-As of 2022-07-02 you also need to use a Rust nightly build.
+[Godot 3.5's prebuilt HTML5 export template is built with Emscripten 3.1.10](https://github.com/godotengine/godot/blob/3.5/.github/workflows/javascript_builds.yml).
+It might be possible to use it if build your Rust code with that exact version, but extra compiler flags may be needed. This guide focuses on building the export template yourself with a recent version of Emscripten.  
+As of 2022-09-04 you also need to use a Rust nightly build.
 
 Confirmed working versions are:
-* rustc 1.64.0-nightly (f2d93935f 2022-07-02)
-* Emscripten 3.1.16-git (e026c0e22a986d92695b2563a1e8ef140b5d8a87)
+* rustc 1.65.0-nightly (8c6ce6b91 2022-09-02)
+* Emscripten 3.1.21-git (3ce1f726b449fd1b90803a6150a881fc8dc711da)
 
 
 ### Installing and configuring Emscripten
@@ -75,8 +78,8 @@ In your project's `config.toml`, add the following:
 [target.wasm32-unknown-emscripten]
 rustflags = [
 	"-Clink-arg=-sSIDE_MODULE=2", # build a side module that Godot can load
-	"-Cpanic=abort", # panic unwinding is currently broken without -sWASM_BIGINT, see below
-	#"-Clink-arg=-sWASM_BIGINT", # alternative to panic=abort, requires building godot with -sWASM_BIGINT also
+	"-Zlink-native-libraries=no", # workaround for a wasm-ld error during linking
+	"-Cpanic=abort", # workaround for a runtime error related to dyncalls
 ]
 ```
 
@@ -91,11 +94,21 @@ cargo +nightly build --target=wasm32-unknown-emscripten --release
 
 The result is a `.wasm` file that you add in your `GDNativeLibrary` properties under `entry/HTML5.wasm32`, just like you would add a `.so` or a `.dll` for Linux or Windows.
 
+### Errors you might encounter
+
+Compile time:
+* `failed to run custom build command for gdnative-sys v0.10.1`, `fatal error: 'wchar.h' file not found`: Emscripten cache not populated, build Godot export template first
+* `undefined symbol: __cxa_is_pointer_type`: You need to build with `-Clink-arg=-sSIDE_MODULE=2`
+
+Runtime:
+* `indirect call signature mismatch`: Possibly due to Emscripten version mismatch between Godot and Rust
+* `need the dylink section to be first`: Same as above
+* `WebAssembly.Module(): Compiling function #1 failed: invalid local index`:  You need to build with `-Cpanic=abort`
+
 ### Further reading
 
-In the future, some/all of this will probably not be needed, for more info see:
+In the future, some of this will probably not be needed, for more info see:
 * [Tracking issue for godot-rust wasm support](https://github.com/godot-rust/godot-rust/issues/647)
+* [Rust](https://github.com/rust-lang/rust/issues/98155) and [Emscripten](https://github.com/rust-lang/rust/pull/98303#issuecomment-1162172132) issues that explain the need for `-Zlink-native-libraries=no`
+* [Emscripten PR that should have obsoleted `-Cpanic=abort`](https://github.com/emscripten-core/emscripten/pull/17328)
 * [Rust PR that would obsolete `-Clink-arg=-sSIDE_MODULE=2`](https://github.com/rust-lang/rust/pull/98358)
-* [Godot PR that updates Emscripten version for Godot 3.x](https://github.com/godotengine/godot/pull/61989)
-* [Godot PR that would obsolete `-Cpanic=abort` in favor of `-sWASM_BIGINT`](https://github.com/godotengine/godot/pull/62397)
-* [Emscripten PR that would obsolete `-Cpanic=abort` without requiring `-sWASM_BIGINT`](https://github.com/emscripten-core/emscripten/pull/17328)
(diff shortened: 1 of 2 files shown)


src/exporting/html5.md
```diff
@@ -0,0 +1,92 @@
+# HTML5
+
+Exporting to HTML5 works just like exporting to other platforms, however there are some things that can make the process a bit tricky and require extra attention.
+
+## General considerations
```
jacobsky: Since this export process doesn't have a trivial set up, I think this section can be called "Tooling Requirements", "Setup Requirements"  or something like that. This would make it more obvious to users that there are very specific requirements for this export process.

src/exporting/html5.md
```diff
@@ -0,0 +1,92 @@
...
+```bash
+emcc -v
+rustc --version --verbose
+```
+
+Also note that `wasm32-unknown-emscripten` is a 32-bit target, which can cause problems with crates incorrectly assuming that `usize` is 64 bits wide.
+
+## Godot 3.4.4
+
+**Disclaimer**: _Currently, the following steps are only tested and confirmed to work on Linux._
```
jacobsky, line 29: It might be worth making this disclaimer into it's own section where we can include all of the additional version requirements. In the future, more folks may be able to test this and confirm working versions so having a section could be useful for new users.

src/exporting/html5.md
```diff
@@ -0,0 +1,92 @@
...
+
+Godot 3.4.4's export template is built with a very old Emscripten version that doesn't play nice with Rust. We will have to build it ourselves with a more recent version.  
+We also need to build the `std` crate ourselves, so we will have to use a nightly Rust build.
+
+Confirmed working versions are:
+* rustc 1.63.0-nightly (dc80ca78b 2022-06-21)
+* Emscripten 3.1.15-git (4100960b42619b28f19baf4254d5db2097234b32)
+
+
+### Getting Emscripten
```
jacobsky: Minor nitpick, "Installing and configuring Emscripten" might be a more appropriate section title.

src/exporting/html5.md
```diff
@@ -0,0 +1,92 @@
...
+./emsdk install tot
+
+# Make the "tot" SDK "active" for the current user. (writes .emscripten file)
+./emsdk activate tot
+
+# Activate PATH and other environment variables in the current terminal
+source ./emsdk_env.sh
+```
+
+### Building Godot
```
jacobsky, line 61: Does this require building Godot from source? or only the export template? If the latter, can we change the title to "Building Godot export template"

src/exporting/html5.md
```diff
@@ -0,0 +1,92 @@
...
+
+Build the Godot Export template according to the [instructions](https://docs.godotengine.org/en/stable/development/compiling/compiling_for_web.html):
+
+```bash
+source "/path/to/emsdk-portable/emsdk_env.sh"
+scons platform=javascript tools=no gdnative_enabled=yes target=release
+mv bin/godot.javascript.opt.gdnative.zip bin/webassembly_gdnative_release.zip
+```
+
+Set the newly built export template as a custom template in Godot and be sure to set the export type as GDNative. When exporting, uncheck "Export With Debug".
```
jacobsky: I'm assuming that the export template requires installing from a file as per: https://docs.godotengine.org/en/stable/tutorials/export/exporting_basics.html#export-templates

Is that correct? If so, can I ask for you to add the above link for completeness?

jacobsky (changes requested):
Thank you so much for the PR! This is a great guide that helps better define how godot-rust supports HTML5 exports.

I have included some minor nitpicks on structure. I will also give this process a go to test this from my end.

@Bromeon feel free to let me know if you have any additional thoughts or questions on this addition

src/exporting/html5.md
derivator: As far as I can tell an export template is just a pre-built version of Godot, so these are the same thing. We are compiling the engine from source to wasm.

src/exporting/html5.md
derivator: This is what I'm referring to: 
image

We could link that image?

src/exporting/html5.md
derivator: I expect this whole section to become outdated very rapidly. What I said under "General Considerations" (matching emscripten versions) is fairly general, but building Godot will probably not be necessary with 3.6, and rust compiler flags could change any second now.

src/exporting/html5.md
Bromeon: Sounds good to me!

src/exporting/html5.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Building your Rust code
+
+In your project's `config.toml`, add the following:
```
Bromeon, line 75: This refers to the godot-rust crate which is compiled as a `cdylib`? If yes, maybe quickly mention that to resolve ambiguities 🙂

src/exporting/html5.md
Bromeon: Since you built Godot from source, did you use `gdnative` with the `custom-godot` Cargo feature?

Bromeon (commented):
This is an awesome addition, thank you very much! ❤️ 

Good that you highlighted that some of the guide might be obsolete soon. I think it's good to start somewhere, and update when needed. Hopefully, a smaller scale in updating also motivates more people to contribute 🙂

derivator:
@Bromeon I don't use `custom-godot`, should I? I'm just rebuilding 3.4.4 with a newer Emscripten, but it's still 3.4.4

Bromeon:
@derivator You should only need `custom-godot` if the generated API JSON would differ from the one shipped in the godot-rust repo. This is usually the case for other Godot minor versions. 

I'm not sure if using Emscripten has an effect here. You could run
```bash
godot --gdnative-generate-json-api api.json
```
and diff the generated file against the `api.json` in the godot-rust repo.

derivator:
> ```shell
> godot --gdnative-generate-json-api api.json
> ```

I have no idea how I would do that for a wasm build, but I suspect it's fine. It works for me after all :laughing:

src/exporting/html5.md
```diff
@@ -0,0 +1,92 @@
...
+
+```toml
+[target.wasm32-unknown-emscripten]
+rustflags = [
+	"-Clink-arg=-sSIDE_MODULE=2", # build a side module that Godot can load
+	"-Crelocation-model=pic", # needed to prevent linker errors
```
hoodmane: This is not needed on nightly Rust:
https://github.com/rust-lang/rust/pull/98149

src/exporting/html5.md
```diff
@@ -0,0 +1,92 @@
...
+	"-Cpanic=abort", # panic unwinding is currently broken without -sWASM_BIGINT, see below
```
hoodmane: PR to fix:
https://github.com/emscripten-core/emscripten/pull/17328

derivator: I've updated the guide to reflect the current state of affairs.

I initially had some problems where I still needed `-Zbuild-std` (to fix missing ` __cxa_is_pointer_type`) and `-Zlink-native-libraries=no` (to fix `wasm-ld: error: unable to find library -lc`, but after `cargo clean` it works without those flags so maybe rust was incorrectly reusing some old build artifact? @hoodmane can you confirm that these flags shouldn't be needed?

hoodmane: Currently Pyodide is using:
```
export RUSTFLAGS= \
	-C link-arg=-sSIDE_MODULE=2 \
	-C link-arg=-sWASM_BIGINT \
	-Z link-native-libraries=no
```
https://github.com/pyodide/pyodide/blob/main/Makefile.envs#L159

The problem that `-Zlink-native-libraries=no` fixes is not resolved either on the Rust side or on the Emscripten side. But if it works without it...

src/exporting/html5.md
jacobsky: Fair point. I think keeping a normal disclaimer is probably enough in that case.

src/exporting/html5.md
```diff
@@ -0,0 +1,101 @@
...
+A version mismatch between Rust and Emscripten will probably result in compiler errors, while incompatible Emscripten versions between Rust and Godot will result in cryptic runtime errors such as `indirect call signature mismatch`.
+
+Also note that `wasm32-unknown-emscripten` is a 32-bit target, which can cause problems with crates incorrectly assuming that `usize` is 64 bits wide.
+
+## Godot 3.4.4
+
+**Disclaimer**: _Currently, the following steps are only tested and confirmed to work on Linux._
+
+Godot 3.4.4's export template is built with a very old Emscripten version that doesn't play nice with Rust. You will have to build it yourself with a more recent version.  
+As of 2022-07-02 you also need to use a Rust nightly build.
```
Bromeon: Could you maybe specify which features of nightly Rust are needed; is it for the `rustflags` below?

Anything the user needs to specify/add in the code, like
```rs
#![feature(some_unstable_thing)]
```

src/exporting/html5.md
```diff
@@ -0,0 +1,101 @@
...
+
+Confirmed working versions are:
+* rustc 1.64.0-nightly (f2d93935f 2022-07-02)
+* Emscripten 3.1.16-git (e026c0e22a986d92695b2563a1e8ef140b5d8a87)
```
Bromeon: This is awesome for reproducibility, thanks a lot! 👍

On which operating system did you test this? Do you think there might be issues on another OS?

Bromeon (commented):
Thanks to you both for the great writeup already! 😊

src/exporting/html5.md
derivator: > Could you maybe specify which features of nightly Rust are needed; is it for the rustflags below?

Previous Rust versions required even more extra flags to get it working, and to build `std` with `-Zbuild-std`, so nightly was required anyway. Thanks to @hoodmane you don't need to build `std` any more on current nightly version, but to my knowledge, it just doesn't work on stable at the moment.

No unstable features are required at the moment.

src/exporting/html5.md
derivator: I tested on linux, and the instructions assume you're using a bash-like shell. I don't see an obvious reason why it *wouldn't* work on Windows (Rust, Emscripten, Godot and SCons all work on Windows after all), but I haven't tested it and have no interest in doing so.

Bromeon: Makes sense, thanks for the clarification! 🙂

src/exporting/html5.md
```diff
@@ -0,0 +1,101 @@
...
+* use a specific version of Rust
+
+You can check which versions of Emscripten/Rust you are using like this:
+
+```bash
+emcc -v
+rustc --version --verbose
+```
+
+A version mismatch between Rust and Emscripten will probably result in compiler errors, while incompatible Emscripten versions between Rust and Godot will result in cryptic runtime errors such as `indirect call signature mismatch`.
```
hoodmane: > will result in cryptic runtime errors such as `indirect call signature mismatch`

It might be worth indicating that `indirect call signature mismatch` does not necessarily mean that there is a version incompatibility: this error message is also the most likely outcome of memory corruption and several other problems.

Bromeon: @derivator any comment on this? 🙂

Bromeon:
@derivator @hoodmane Any more comments before this is merged? 🙂

derivator:
@Bromeon As far as I can tell at a glance, building Godot shouldn't be necessary anymore with Godot 3.5, as the commit that updates their emscripten version seems to have made it in, but I would still keep that information in the guide, in case the same kind of version mismatch happens in the future. I can try updating the guide this weekend.

Bromeon:
@derivator have you had a chance to look at it yet? 🙂

derivator:
@Bromeon I'm having a look right now. Some steps will not be necessary anymore, however [Emscripten makes *no* ABI stability promises](https://github.com/emscripten-core/emscripten/blob/main/ChangeLog.md), so while using the export template provided by godot *might* work coincidentally, I think it's probably best to always compile it yourself. I will update the guide to reflect this.

derivator:
I have it working with Godot 3.5 following the steps in the guide. I believe `-Cpanic=abort` shouldn't be necessary anymore thanks to https://github.com/emscripten-core/emscripten/pull/17328, but I'm still getting errors without it.

Bromeon:
Thanks a lot to everyone involved! It's great to finally have a bit more resources on the topic of HTML5 exports ❤️

GeTechG:
I was able to run on emscripten 3.1.14, Godot 3.5.1, rust nightly 0.65.0 (2022-09-03). I was unable to run as suggested on the latest versions of rust and emscripten, or even 3.1.21 as in the tutorial. A runtime error came out with similar text: 
```
at: init_library (modules/gdnative/nativescript/nativescript.cpp:1479) - Condition "lib_path.length() == 0" is true.
```
all because I understood different versions of llvm, so I am puzzled how you could run with the latest version of emscripten, as nightly rust builds still have 15.0, not 16.0.

Issue #72, opened 2022-06-26
Godot property hooks documentation is out of date


jacobsky:
[Property access hooks](https://godot-rust.github.io/book/rust-binding/properties.html#property-access-hooks) is no longer valid. As per https://github.com/godot-rust/godot-rust/pull/874

The new way to register is with the with `#[property(get|set)]`

Bromeon:
Closed by #73.

PR #73, opened 2022-07-06
Updating the Outdated API document as per issue #72

jacobsky:
This should resolve issue #72 as well as better illustrate the options for getters and setters with Godot Rust

The code that is in the document has all been validated in my own test project.

### DIFF
--- a/src/rust-binding/properties.md
+++ b/src/rust-binding/properties.md
@@ -49,6 +49,7 @@ Default get/set functions can be registered as per the following example:
 #[inherit(Node)]
 struct GodotApi {
     // property registration
+    // Note: This is actually equivalent to #[property]
     #[property(get, set)]
     prop: i32,
 }
@@ -104,7 +105,7 @@ struct GodotApi {
 
 impl GodotApi {
     fn new(_owner: &Node) -> Self {
-        GodotApi { prop: "".to_owned() }
+        GodotApi { prop: String::new() }
     }
 }
 
@@ -196,8 +197,7 @@ impl MyNode {
 
 Sometimes it can be useful to expose a value as a property instead of as a function. Properties of this type serve as a `marker` for the existence of a property and allows it to be viewed in the editor, even if it doesn't exist by itself.
 
-This can be useful for data (similar to the first sample) where you want to have a property visible in the editor without having to have a concrete data type.
-For example in the following:
+This can be useful for data (similar to the first sample) where the count serves more as a property of `enemies` rather than as it's own distinct data, such as the following: 
 
 ```rs
 struct Enemy {
@@ -206,13 +206,14 @@ struct Enemy {
 #[derive(NativeClass)]
 struct GodotApi {
     enemies: Vec<Enemy>,
+    // Note: As the property is a "marker" property, this will never be used in code.
     #[allow(dead_code)]
     #[property(get = "Self::get_size")]
     enemy_count: Property<u32>,
 }
 
 #[methods]
-impl MyVec {
+impl GodotApi {
     //...
 
     fn get_size(&self, _owner: TRef<Reference>) -> u32 {
--- a/src/rust-binding/properties.md
+++ b/src/rust-binding/properties.md
@@ -37,52 +37,87 @@ If you need to hide this property in Godot editor, use `no_editor` option:
 enemy_count: i32,
 ```
 
-## Property access hooks
+## Property get/set
 
-You can add hooks to call methods before and after property value was retrieved or changed.
+Properties can register `set` and `get` methods to be called from Godot.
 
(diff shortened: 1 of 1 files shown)


src/rust-binding/properties.md
```diff
@@ -37,52 +37,87 @@ If you need to hide this property in Godot editor, use `no_editor` option:
...
-    // before get
-    #[property(before_get = "Self::before_get")]
-    prop_before_get: i32,
-
-    // before set
-    #[property(before_set = "Self::before_set")]
-    prop_before_set: i32,
+    // property registration
+    #[property(get, set)]
+    prop: i32,
```
Bromeon: Maybe mention that this specific case is equivalent to a simple `#[property]`.

src/rust-binding/properties.md
```diff
@@ -37,52 +37,87 @@ If you need to hide this property in Godot editor, use `no_editor` option:
...
+    // property registration
+    #[property(get_ref = "Self::get", set = "Self::set")]
+    prop: String,
+}
 
-    fn after_get(&self, _owner: TRef<Node>) {
-        godot_print!("After get");
+impl GodotApi {
+    fn new(_owner: &Node) -> Self {
+        GodotApi { prop: "".to_owned() }
```
Bromeon: Maybe `String::new()` instead of `"".to_owned()`?

It's a bit annoying that there are so many conventions for this in Rust (`to_string()`, `to_owned()`, `into()`, `String::from()`, etc), although I understand they're needed for genericity.

src/rust-binding/properties.md
```diff
@@ -154,3 +191,32 @@ impl MyNode {
     }
 }
 ```
+
+## `Property<T>` and when to use it
+
+Sometimes it can be useful to expose a value as a property instead of as a function. Properties of this type serve as a `marker` for the existence of a property and allows it to be viewed in the editor, even if it doesn't exist by itself.
```
Bromeon: `marker` should probably be italics: _marker_ (it's not a code symbol)

"...and **allow** it to be viewed..." (3rd person plural; subject is "Properties of this type")

"even if it doesn't exist by itself" -- arguably it _does_ exist from a Godot point of view 😉 
Maybe it's clearer to say something like "even if it doesn't exist as a Rust field" or so.

src/rust-binding/properties.md
```diff
@@ -154,3 +191,32 @@ impl MyNode {
...
+
+This can be useful for data (similar to the first sample) where you want to have a property visible in the editor without having to have a concrete data type.
+For example in the following:
```
Bromeon: Not sure I understand this; after all there _is_ a concrete data type (`u32`)...?

Did you mean without a concrete field? In that case I think the paragraph is redundant, as this is already mentioned before.

src/rust-binding/properties.md
```diff
@@ -154,3 +191,32 @@ impl MyNode {
...
+
+```rs
+struct Enemy {
+    // Enemy Data
+}
+#[derive(NativeClass)]
+struct GodotApi {
+    enemies: Vec<Enemy>,
+    #[allow(dead_code)]
```
Bromeon, line 210: Is it worth pointing out why `#[allow(dead_code)]` can be necessary?

Bromeon (commented):
Thanks a lot! 🙂
Made some comments.

src/rust-binding/properties.md
jacobsky: It's the same? I thought that it wouldn't add the set/get by default. That is a surprise 🤣 I will make a note.

src/rust-binding/properties.md
jacobsky: Alrighty. I'm cool with that. (I always initialize via Strings as `"".to_owned()`, I'm not sure when I got into that habit :rofl:)

src/rust-binding/properties.md
jacobsky: Yeah, this is confusing. I think I am trying to say something about "count is a property of the enemy Vector instead of it's own distinct value" or something like that. I'm taking another stab at it.

src/rust-binding/properties.md
jacobsky: Actually, I took that from the tests and hadn't thought about why it was there. I think mentioning why it can be useful would be a good change. I will add a comment.

jacobsky: I also just realized that I was missing the impl block for it. I'll fix that too.

src/rust-binding/properties.md
```diff
@@ -196,8 +197,7 @@ impl MyNode {
 
 Sometimes it can be useful to expose a value as a property instead of as a function. Properties of this type serve as a `marker` for the existence of a property and allows it to be viewed in the editor, even if it doesn't exist by itself.
 
-This can be useful for data (similar to the first sample) where you want to have a property visible in the editor without having to have a concrete data type.
-For example in the following:
+This can be useful for data (similar to the first sample) where the count serves more as a property of `enemies` rather than as it's own distinct data, such as the following:
```
Bromeon: "**its** own" 😉 

This and the above comment (line 198), then should be good!

jacobsky: What an embarrassing typo 🤣

Bromeon:
Thanks a lot! 🚀

Issue #57, opened 2021-11-27
Documentation on rust-analyzer seems outdated / anything to improve integration?


bluenote10:
The section [How do I get auto-completion with rust-analyzer?](https://godot-rust.github.io/book/faq/configuration.html#how-do-i-get-auto-completion-with-rust-analyzer) seems outdated, because [`rust-analyzer.cargo.loadOutDirsFromCheck` no longer exists](https://github.com/rust-analyzer/rust-analyzer/issues/6448#issuecomment-961403175).

In practice this is impacting my productivity quite a lot, because auto-completion within a native class exported function doesn't seem to work at all. Often, my IDE is in a state reporting simply "all broken"

![image](https://user-images.githubusercontent.com/3620703/143691057-fe1b00a1-c569-40ad-aa52-b5fc8d11dca5.png)

... and if this happens when I'm having a weak moment it can take a while to spot the actual error.

To avoid getting impacted by this issue I have now started to implement more complex methods as free floating functions like
```rust
fn process(this: &mut MyClass, delta: f64) -> Option<()> {
    // ...
}
```
and forward calls from the methods, which is of course unnecessary boilerplate.

------

To narrow down the problem we probably need to answer the questions:

1. Is this just something broken with my setup, or is everyone using rust-analyzer facing these issues?

2. If it is setup related, what are the tricks to get it going, which we could update in the book?

3. If it turns out that it is fundamentally not working, the question may be if we can do something on library / macro side to improve it. Following [the links posted](https://github.com/rust-analyzer/rust-analyzer/issues/6448#issuecomment-961441435) by the rust-analyzer maintainer leads to a [similar issue](https://github.com/rust-analyzer/rust-analyzer/issues/10520) for `tokio-tracing`, caused by the usage of `syn` and [this proposal](https://github.com/tokio-rs/tracing/issues/1633) to solve it on their side. @Bromeon Is this proposal applicable to godot-rust as well?

jacobsky:
@bluenote10 I agree, that part of the FAQ appears to be outdated. Thanks for creating the issue!

I have noticed similar issues occurring with the recent changes to Rust Analyzer, a lot of the issues appear to be related to rust-analyzer's macro expansion, so I ended up disabling them for the time being. I haven't noticed too many issues with rust analyzer though. It would be good to see what other folks using Rust Analyzer experience as well.

Bromeon:
Does anyone who uses Rust-Analyzer have time to investigate this issue?
[...]

PR #74, opened 2022-07-13
Removes FAQ entries related to Rust Analyzer as they no longer appear to be relevant.

jacobsky:
I tested it on a fresh install of rust-analyzer and it doesn't seem to have any issues at the moment. If anyone finds anything wrong, please comment here

Closes issue #57 as rust analyzer does not seem to have issue with code completion.


jacobsky:
As there appear to be no objections after this week, I will go ahead and merge this issue in. It's pretty small change.

PR #75, opened 2022-07-18
Changed signal declaration, so it fits the current docs (0.10.0)

tuto193:
Updated the shown signal registration, so it uses the correct/actual syntax (as seen in the docs).

### DIFF
--- a/src/faq/code.md
+++ b/src/faq/code.md
@@ -64,7 +64,7 @@ struct SignalEmitter {
 #[methods]
 impl SignalEmitter {
     fn register_signals(builder: &ClassBuilder<Self>) {
-        builder.add_signal(Signal { name: "updated", args: &[] });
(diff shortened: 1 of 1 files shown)


jacobsky:
Thank you for the PR fixing this issue. It looks good to me!

One question, did you notice any other places in the book where the signal registration was wrong? Or was it just this one faq entry?

tuto193:
> Thank you for the PR fixing this issue. It looks good to me!
> 
> One question, did you notice any other places in the book where the signal registration was wrong? Or was it just this one faq entry?

image
That seemed to be the only place using the outdated syntax after `grep`ing over the whole book :)

jacobsky:
@tuto193 Thank you so much for checking! I'll go ahead and merge this change. Thank you for your contribution :)

PR #76, opened 2022-07-18
FAQ: Added unwraping after `cast` call, so it compiles correctly.

tuto193:
The `onready var FAQ` had a missing `unwrap`, which didn't allow the code to compile correctly.
Also adjusted `struct` declaration, so `NativeClass` actually uses the `Option<Instance<T>>` format.

This is the only instance, where that is missing in the book FYI (picture attached)

image


Bromeon:
Thanks a lot! 🙂

PR #80, opened 2022-09-19
Update Nix build system instructions

Shou:
We account for some missing headers by documenting how to include them and also documenting how to handle any future missing headers.


Bromeon:
Thanks a lot! 👍 

Closes https://github.com/godot-rust/godot-rust/issues/945.

PR #84, opened 2022-11-11
Update doc links that go to a 404

tonyfinn:
It seems the package was rearranged in 0.10.0 but the docs were not updated so the links to `Ref`, `TRef` and `Instance` lead to a 404 error. This updates them to point to the 0.10.0+ locations for these types.


Bromeon:
Thanks a lot!

PR #87, opened 2022-12-10
Added tutorial for `HTML5 web exports without recompiling Godot`

greenfox1505:
Added a new tutorial for exporting to HTML5, but without needed to recompile Godot itself, by being more explicit with Emscripten version.

I tried to merge with @derivator 's original tutorial, but the reading flow was so different I just decided to add it to a new file. The old tutorial has a lot of notes that I don't, so perhaps it would be better to merge their tutorial into mine. It seems they know much better **why** various compile flags are added. Where as I only know how to copypaste.

I expect this will need another revision before merge.

### DIFF
--- a/src/gdnative/export/easier_html5.md
+++ b/src/gdnative/export/easier_html5.md
@@ -1,0 +1,200 @@
+# HTML5 Exports without Recompiling the Engine
+
+## Limitations and Expectations
+
+Before you begin this tutorial, I expect you:
+
+- Have a meaningful grasp of Godot
+- Have a working understanding Rust
+- Have a Godot3.5.1+Rust project that works for **Linux** builds.
+- That project doesn't use threads or can be made to not use threads.
+- Have access to a computer running Ubuntu 20.04, a Docker runner, a Virtual Machine.
+
+You can might get a way with other Ubuntu versions, but Godot3.5.1 built their export templates with an Ubuntu 20.04 Docker Container, so I'll use 20.04 for this tutorial.
+
+## Setup Environment
+
+You have two options. If your build environment uses Docker, you can just use my `dockerfile` included at the [bottom of this tutorial](#Docker). Build that image and start a container with your project mounted. Then you can skip to [#Building](#building).
+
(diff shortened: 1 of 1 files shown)


Bromeon:
Thanks a lot!

Is this still relevant now that https://github.com/godot-rust/book/pull/89 has been merged? It looks like that PR is a bit more deterministic and has more of a "tutorial" rather than "recipe" format. The problem with WASM in general is that the setup is rather brittle, so we should rather focus on one manual (and make it good) than multiple competing ones.

Do you have certain additions you'd still like to include on top of #89?

greenfox1505:
I haven't looked into this since migrating Godot 4, which is missing other key features for web exports anyway. When I wrote this, I just wanted to document my findings and didn't want to stomp over exiting docs. I agree it's probably better to just have one document, but I didn't want to overwrite someone else's work with an unsolicited tutorial. Didn't go noticed for 5months, so I stopped paying attention too. Not sure I know the difference between a "recipe" and a "tutorial"?

I think it's still relivent today and don't see how the changes on #89 would change that. The tutorial, as it stands today says:

> [Godot 3.5's prebuilt HTML5 export template is built with Emscripten 3.1.10](https://github.com/godotengine/godot/blob/3.5/.github/workflows/javascript_builds.yml). It might be possible to use it if build your Rust code with that exact version, but extra compiler flags may be needed. This guide focuses on building the export template yourself with a recent version of Emscripten.

#89 did not change this text and still tutorializes under the assumption that you are using the latest Emscripten and building your own export templates. The entire point of this tutorial was to show that building your own export templates is not strictly necessary (if you need newer Emscripten features, I suppose it might be). However, I haven't been in Godot-Rust for a while, so it's possible an update broke this tutorial.

These two tutorials (recipes?) could be merge into one, branching tutorial. But I did demonstrate how to build in a way isn't demonstrated in the version currently up.

(sorry for late response, for a while gmail was eating my github notifications, so I missed this and only found it again when I was curious the state of Godot4+Rust WASM)

PR #89, opened 2023-01-28
Update HTML5 export guide with pinned versions of Rust and Emscripten

### LINKED ISSUE #88: Commands in the WASM guide should use pinned versions of rustc and emsdk
It appears that the versions of LLVM used in rustc and emscripten have drifted out of sync again. Although the guide does specify a pair of known good versions of the tools, the commands simply use `nightly` and `tot` instead, and as a result some linking issues may arise when commands are blindly copied.

There are several steps we can take here. Firstly, the commands should be modified to use pinned versions. Longer term, we should see if there's a consistent way to tell the compatible version of emscripten from any given rustc version, or if we have to stick to manually finding known good pairs for the foreseeable future.

We should also see if #87 is still viable and if it can be merged into the current guide.

necrashter:
Closes #88, among other minor changes.

Added a list of compatible Rust and Emscripten versions, updated the commands according to the latest known working combination.

Removed "A version mismatch between Rust and Emscripten will probably result in compiler errors, while using incompatible Emscripten versions between Rust and Godot will result in cryptic runtime errors.", which was misleading for my case. I had a version mismatch but I was getting the runtime error `ERROR: Can't resolve symbol godot_gdnative_init`. I added this error to the list of known errors.

### DIFF
--- a/src/gdnative/export/html5.md
+++ b/src/gdnative/export/html5.md
@@ -4,9 +4,10 @@ Exporting to HTML5 works just like exporting to other platforms, however there a
 
 ## What you need (to know)
 
-The Godot HTML5 export templates are built with [Emscripten](https://emscripten.org/), so you need to build your Rust code for the `wasm32-unknown-emscripten` target (`wasm32-unknown-unknown` will not work).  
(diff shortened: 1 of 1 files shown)


chitoyuu (approved):

chitoyuu:
Corrected information around LLVM version reporting by `emcc`, did some slight formatting changes and merged. Thanks!

Issue #81, opened 2022-09-26
Nix install instructions use outdated/unsupported 1.52.1 compiler


kpcyrd:
Somebody in ##rust on irc had issues due to using a pre-2021 edition of rustc/cargo, turned out they were following the [Recipe: Nix as development environment](https://godot-rust.github.io/book/recipes/nix-build-system.html#recipe-nix-as-development-environment) chapter of the godot-rust book which pins a specific version of nixpkgs:

```
let
    # Get an up-to-date package for enabling OpenGL support in Nix
    nixgl = import (fetchTarball "https://github.com/guibou/nixGL/archive/master.tar.gz") {};

    # Pin the version of the nix package repository that has Godot 3.2.3 and compatible with godot-rust 0.9.3
    # You might want to update the commit hash into the one that have your desired version of Godot
    # You could search for the commit hash of a particular package by using this website https://lazamar.co.uk/nix-versions
    pkgs = import (fetchTarball "https://github.com/nixos/nixpkgs/archive/5658fadedb748cb0bdbcb569a53bd6065a5704a9.tar.gz") {};
in
    # Configure the dependency of your shell
    # Add support for clang for bindgen in godot-rust
    pkgs.mkShell.override { stdenv = pkgs.clangStdenv; } {
        buildInputs = [
            # Rust related dependencies
            pkgs.rustc
            pkgs.cargo
            pkgs.rustfmt
            pkgs.libclang

            # Godot Engine Editor
            pkgs.godot

            # The support for OpenGL in Nix
            nixgl.nixGLDefault
        ];

        # Point bindgen to where the clang library would be
        LIBCLANG_PATH = "${pkgs.libclang.lib}/lib";
        # Make clang aware of a few headers (stdbool.h, wchar.h)
        BINDGEN_EXTRA_CLANG_ARGS = with pkgs; ''
          -isystem ${llvmPackages.libclang.lib}/lib/clang/${lib.getVersion clang}/include
          -isystem ${llvmPackages.libclang.out}/lib/clang/${lib.getVersion clang}/include
          -isystem ${glibc.dev}/include
        '';

        # For Rust language server and rust-analyzer
        RUST_SRC_PATH = "${pkgs.rust.packages.stable.rustPlatform.rustLibSrc}";

        # Alias the godot engine to use nixGL
        shellHook = ''
            alias godot="nixGL godot -e"
        '';
    }
```

[...]

PR #92, opened 2023-03-01
Fix nix page

JasmineLowen:
I noticed, that one of the packages used in the nix shell example doesn't work
anymore. I made the necessary changes and with this it builds again on my
machine.

### DIFF
--- a/src/gdnative/recipes/nix-build-system.md
+++ b/src/gdnative/recipes/nix-build-system.md
@@ -37,7 +37,7 @@ in
             pkgs.godot
 
             # The support for OpenGL in Nix
-            nixgl.nixGLDefault
+            nixgl.auto.nixGLDefault
         ];
 
         # Point bindgen to where the clang library would be


chitoyuu (approved):
Thanks for the PR! I'm not familiar with Nix at all, but it does appear that this is the currently recommended way to install `nixgl` according to its [README](https://github.com/guibou/nixGL/blob/c917918ab9ebeee27b0dd657263d3f57ba6bb8ad/README.md):

> OpenGL wrappers:
>
> `auto.nixGLDefault`: Tries to auto-detect and install Nvidia, if not, fallback to mesa. Recommended. Invoke with `nixGL program`.

While it'd be best if another Nix user can come in and help verify the solution, this change does look reasonable, so I'll be merging it without blocking it further. For future PRs like this, please try to cite some sources, so they're easier for us to review.

Bromeon:
@RobWalt Thanks a lot for this!

If you have the time, could you maybe also have a look at Nix-related issue #81 and tell us what you think? 🙂

JasmineLowen:
@Bromeon I already had a look over it. It works fine for me and I don't really get what the problem is/was from the description that was provided. I'll try to resolve the issue

PR #94, opened 2023-03-02
Elaborate GDNative Explanation

JasmineLowen:
I know that Godot 4 just released and that GDNative will probably play less of a role for the future, but I wanted to update the old page here since I had some trouble finding all the right places to configure stuff.

I odn't know if you people accept commits including images. If that isn't the case I can try to create a more detailed guide with a few more paragraphs instead.


chitoyuu (changes requested):
Thanks for the PR! Screenshots should be very helpful for new users. It would be great however if they could be placed more locally related to the chapter in which they are used, e.g. `/src/gdnative/intro/img/*.png`, like how it's done for existing chapters with images in them. This way, it's much easier for us to maintain in case more images are added for different parts of the book in the future.

JasmineLowen:
Makes sense! Done! 👍🏼

chitoyuu:
Sorry I let this slip out of my mind for a while. Looks good to me and merged. Thanks!

PR #96, opened 2023-03-04
Update Android Export Section: Android NDK version, troubleshooting, custom Godot

### LINKED ISSUE #95: Android NDK version >= 23 requires nightly toolchain
`libgcc` was removed in Android NDK version `r23-beta3` and Rust was updated accordingly (see [this issue](https://github.com/rust-lang/rust/pull/85806)). However, this change does not seem to be available in stable Rust yet (2023-03-04). This is should be mentioned in [exporting for Android](https://godot-rust.github.io/book/gdnative/export/android.html) section.

I will make a PR for this but I'm not sure where we should mention it. My idea is to add a "Troubleshooting" section similar to ["Errors you might encounter"](https://godot-rust.github.io/book/gdnative/export/html5.html#errors-you-might-encounter) in HTML5 export page. But I think it should be included at the start as well since users start following instructions before reading the troubleshooting.

necrashter:
Closes #95, adds info about Android NDK versions, custom Godot build, and troubleshooting. Also includes some minor corrections.

### DIFF
--- a/src/gdnative/export/android.md
+++ b/src/gdnative/export/android.md
@@ -9,6 +9,21 @@ In order to export to Android, we need to compile our Rust source for the approp
 
 First, we need to install the **Android SDK** with **NDK** enabled. This contains the necessary tools for each architecture. Once the Android SDK is installed, open Editor Settings in the Godot GUI (*Editor > Editor Settings > Export > Android*) and set the **absolute paths** to `adb`, `jarsigner`, and the debug keystore (`debug.keystore`), all of which can be found in the Android SDK installation.
 
+> ### About NDK versions
+>
+> `libgcc` was removed in Android NDK version `r23-beta3`. Although Rust was updated accordingly (see [this issue](https://github.com/rust-lang/rust/pull/85806)), it's not available in stable Rust yet (as of 2023-03-04). Therefore, you need to use the **nightly toolchain** if you have **Android NDK version 23 or higher**.
+>
+> ```bash
+> # Install nightly toolchain
+> rustup toolchain install nightly
+> ```
+>
+> After that, run all `rustup` or `cargo` commands in this tutorial using the nightly toolchain by adding `+nightly` argument. For example:
+>
+> ```bash
+> rustup +nightly target add aarch64-linux-android
+> ```
+
 Then, we'll install the Rust toolchains for the targets we want to support:
 
 ```bash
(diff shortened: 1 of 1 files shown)


src/gdnative/export/android.md
```diff
@@ -37,7 +52,7 @@ rustup target add armv7-linux-androideabi  # for armv7 (32-bit)
 rustup target add i686-linux-android       # for x86 (32-bit)
 ```
 
-#### `gcc` libraries for cross-compilation
+### `gcc` libraries for cross-compilation
```
necrashter: This is not a subsection of "32-bit targets" header. I thought it was when I first read it.

chitoyuu: It's not? Are the 32-bit versions of gcc also necessary if you're just compiling for a 64-bit target?

src/gdnative/export/android.md
```diff
@@ -9,6 +9,21 @@ In order to export to Android, we need to compile our Rust source for the approp
...
+>
+> ```bash
+> # Install nightly toolchain
+> rustup toolchain install nightly
+> ```
+>
+> After that, run all `rustup` or `cargo` commands in this tutorial using the nightly toolchain by adding `+nightly` argument. For example:
+>
+> ```bash
+> rustup +nightly target add aarch64-linux-android
```
chitoyuu, line 24: Maybe it's worth it to also mention the `rust-toolchain.toml` configuration file, as a more persistent way to configure the toolchain for a project? https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file

src/gdnative/export/android.md
necrashter: I couldn't compile for `aarch64-linux-android` target without installing the given packages. I think `g++-multilib` and `gcc-multilib` are required to compile for ARM64 from a x86_64 machine. `libc6-dev-i386` may not be necessary though. I will try removing the packages and check which ones are necessary.

src/gdnative/export/android.md
necrashter: I didn't know about that, thanks! I personally use nightly for HTML5 and Android, and stable for desktop. But I will mention the link you've provided in the next update.

src/gdnative/export/android.md
necrashter: You are right, none of them are required for `aarch64-linux-android` target. I will revert this.

chitoyuu (approved):
Looks good to me. Thanks!

PR #97, opened 2023-03-14
Gdextension

Local-Trash:
Added a startup for gdextensions.


src/SUMMARY.md
```diff
@@ -1,7 +1,10 @@
 # Summary
 
 - [Introduction](index.md)
-- [GDExtension](gdextension/index.md)
+- [GDExtension](gdext/index.md)
+  - [Getting Started](gdext/getting%20started/index.md)
+    - [Setup](gdext/getting%20started/setup.md)
+    - [Coding](gdext/getting%20started/coding.md)
```
Bromeon: I'd try to avoid spaces in URLs. Could you adopt the `intro` convention from gdnative?
```
gdext/intro/index.md
gdext/intro/setup.md
gdext/intro/coding.md
```

src/gdext/getting started/coding.md
```diff
@@ -0,0 +1,69 @@
...
+use godot::prelude::*;
+
+struct Main;
+
+#[gdextension]
+unsafe impl ExtensionLibrary for Main {}
+```
+
+## Creating a Node Type
+To control a node you need to create a node type that inherits the node type. For an example, Let's use a CharacterController2D.
```
Bromeon: > To control a node you need to create a node type that inherits the node type.

That sounds strange. Maybe:

> To control a node of a given type, you can define a Rust struct that "inherits" that type. [then proceed quickly mentioning inheritance, as it's a foreign concept in Rust]

src/gdext/getting started/coding.md
```diff
@@ -0,0 +1,69 @@
...
+struct Player {
+    speed: f32,
+    gravity: f32,
+
+    #[base]
+    base: Base<CharacterController2D>
+}
+```
+Let's break this down. The derive allows gdextensions to use the struct and turn it into a godot class. The class attribute allows you customize the Node. In this case we made the base Character Controller 2D. Then, in the struct speed and gravity fields are not important just help with player speed. The base field lets edextensions know where to store the Node data. 
+
```
Bromeon: Terminology, spacing and capitalization -- please also use single backticks for code symbols.
``` 
gdextensions             -> gdext
Character Controller 2D  -> `CharacterController2D`  (note the backticks)
speed and gravity        -> `speed` and `gravity`
Node                     -> node  (if talking about the concept, not the symbol)
```

src/gdext/getting started/coding.md
```diff
@@ -0,0 +1,69 @@
...
+Now let's add some functionality.
+```rust
+impl GodotExt for Player {
+    fn init(base: Base<CharacterController2D>) -> Self {
+        Self {
+            speed: 5f32,
+            gravity: 5f32,
+
+            base
+        }
```
Bromeon: I would just use `5.0` here, which is more common. Also no need for the empty line between fields.

src/gdext/index.md
```diff
@@ -37,6 +37,12 @@ That said, there are some notable differences:
    
    There is no differentiation between "tool" and "normal" scripts anymore; Rust logic runs as soon as the Godot editor launches.
    godot-rust will likely add an opt-out mechanism to limit execution to the launched game, as that's usually the intended behavior.
+   Currently, if you want to only run when the game runs write this code in you process functions
```
Bromeon:
```suggestion
   Currently, if you want logic to execute only when the game runs, write this code at the beginning of your process functions:
```

Bromeon (commented):
Thanks a lot, this is a very nice first intro! 🙂 

I added a few inline comments. Please also consider capitalization and use consistent naming for different technologies:
```
Rust
Godot
gdext (this library)
GDExtension (the Godot C binding, likely not worth mentioning in the intro)
```

Local-Trash:
I fixed all of the conflicts. I think. If there are any more please tell me.

src/gdext/index.md
```diff
@@ -37,6 +37,12 @@ That said, there are some notable differences:
    
    There is no differentiation between "tool" and "normal" scripts anymore; Rust logic runs as soon as the Godot editor launches.
    godot-rust will likely add an opt-out mechanism to limit execution to the launched game, as that's usually the intended behavior.
+   Currently, if you want logic to execute only when the game runs, write this code at the beginning of your process functions:
+   ```rust
+   if Engine::singleton().is_editor_hint() {
+      return
```
Bromeon, line 43: Missing semicolon.

src/gdext/intro/coding.md
```diff
@@ -0,0 +1,69 @@
+# Coding
+This is where you can write the rust logic.
```
Bromeon: rust -> Rust

src/gdext/intro/coding.md
```diff
@@ -0,0 +1,69 @@
...
+use godot::prelude::*;
+
+struct Main;
+
+#[gdextension]
+unsafe impl ExtensionLibrary for Main {}
+```
+
+## Creating a Node Type
+To add rust functionality to the node. We need to use a base which we can build on. In this case, the base is `CharacterController2D`. To use it we must inherit it. Rust doesn't natively support inherits, but with Godot being written in C++ it uses inherits. Gdext does have a way to use this. By using the `class` attribute.
```
Bromeon:
```suggestion
To add Rust functionality to the node, we need to use a base class which we can build on. In this case, the base is `CharacterController2D`. To use it, we must "inherit" it. Rust doesn't natively support inheritance, but gdext has a way to emulate the Godot inheritance hierarchy: by using the `#[class]` attribute. 
```

src/gdext/intro/coding.md
```diff
@@ -0,0 +1,69 @@
...
+#[class(base=CharacterController2D)]
+struct Player {
+    speed: f32,
+    gravity: f32,
+
+    #[base]
+    base: Base<CharacterController2D>
+}
+```
+Let's break this down. As I mentioned early, we need the `class` attribute, but what does it do? The attribute allows use to store the inherited functions and values in a flied we can use to interact with Godot. If you didn't understand then think of it as it was a field in a struct that holds more than just some rust functions and value. The derive allows gdext to use the struct and turn it into a godot class. The class attribute allows you customize the node. In this case we made the base `CharacterController2D`. Then, in the struct `speed` and `gravity` fields are not important just help with player speed.
```
Bromeon: Try to avoid personal pronouns (and if so, use "We"). Many times, passive is better:

"As I mentioned early" -> "As mentioned earlier"

---

"flied"? Did you mean "file"?

---

> If you didn't understand then think of it as it was a field in a struct that holds more than just some rust functions and value.

I'm not sure what this is clarifying; I think it can be omitted.

---

> Then, in the struct `speed` and `gravity` fields are not important just help with player speed.

Mentioning that they're not important is more confusing here, after all they're part of the state.

---

Again, smaller things:
* `class` -> `#[class]`
* class -> `#[class]`
* godot -> Godot
* allows you -> allows you to

src/gdext/intro/coding.md
```diff
@@ -0,0 +1,69 @@
...
+
+Now let's add some functionality.
+```rust
+impl GodotExt for Player {
+    fn init(base: Base<CharacterController2D>) -> Self {
+        Self {
+            speed: 5.0,
+            gravity: 5.0,
+
```
Bromeon: see my previous comment about newlines

src/gdext/intro/coding.md
```diff
@@ -0,0 +1,69 @@
...
+            base
+        }
+    }
+
+    fn physics_process(&mut self, delta: f64) {
+        if Engine::singleton().is_hint_editor() {
+            return
```
Bromeon, line 59: semicolon

src/gdext/intro/coding.md
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        self.base.set_velocity(Vector2::new(speed, gravity));
+        self.base.move_slide();
```
Bromeon: The method is called `move_and_slide`.

src/gdext/intro/coding.md
```diff
@@ -0,0 +1,69 @@
...
+    }
+}
+```
+This code starts by setting some functions that godot will run. The init function is for Godot to create our Node and the physics_process function is what godot runs every frame. The if statement checks if the game is running and doesn't run in the editor. The set_velocity setst eh velocity of the node and the move and slide moves with collisions.
```
Bromeon: Capitalization and orthography.

src/gdext/intro/setup.md
```diff
@@ -0,0 +1,45 @@
...
+The godot-rust bindings depend on `bindgen`, which in turn [depends on LLVM](https://rust-lang.github.io/rust-bindgen/requirements.html). You may download LLVM binaries from [https://releases.llvm.org/](https://releases.llvm.org/).
+
+After installation, check that LLVM was installed properly:
+
+```bash
+# Check if Clang is installed and registered in PATH
+clang -v
+```
+
+`bindgen` may complain about a missing `llvm-config` binary, but it is not actually required to build the `gdnative` crate. If you see a warning about `llvm-config` and a failed build, it's likely that you're having a different problem!
```
Bromeon: Wrong crate name -- and I wouldn't mention the crate but rather the library:

> required to build the gdext library.

Bromeon (commented):
Thanks for the update!

There are still lots of smaller issues, some of which I mentioned before. For the next time, I would appreciate if you could approach them more systematically, e.g.:
* make sure you applied all comments from the previous review (you can mark them "resolved")
* use search&replace for capitalization issues
* use a spellchecker for orthography and grammar
* try to run the Rust code

This would help me focus more on semantic issues, and would speed up the overall review process. Other than that it's really nice to have a first intro for gdext, thanks a lot for the initiative! 🙂

Bromeon:
Thank you! 🙂

PR #104, opened 2023-05-31
Cleared up a misconception on what deferred calls do

L4Vo5:
The godot docs have contradictory explanations, but deferred calls get pushed to a queue that gets flushed both at the end of physics and idle frames. when it gets flushed, all pending calls are called until it's empty (so calling something as deferred from something that's called as deferred will still happen within the same frame, no matter how many times it recurses)
(strangely, in godot 3 it seems to also flush before idle frames, while in godot 4 it flushes before physics frames. not sure why either is done)


Bromeon:
Thanks a lot! I hope we remember this when writing the Godot 4 equivalents 🙂

PR #109, opened 2023-07-30
Note added for https://github.com/godot-rust/gdext/pull/365

RefinedDev:
- Note added for [No longer run Rust classes inside Godot editor](https://github.com/godot-rust/gdext/pull/365)
- Wrapped the note which talked about no hot-reloading for some platforms into a ~warning~ tip


src/gdext/intro/hello-world.md
```diff
@@ -230,14 +230,16 @@ This is a point where you can compile your code, launch Godot and see the result
...
+```admonish warning
+ **Launching the Godot application**
+ 
+While it's possible to open the Godot editor and press the launch button every time you made a change in Rust, this is not the most
+efficient workflow. Unfortunately there is [a GDExtension limitation][no-reload] that prevents recompilation while the editor is open 
+ (at least on Windows systems -- it tends to work better on Linux and macOS).
+ 
+ However, if you don't need to modify anything in the editor itself, you can launch Godot from the command line or even your IDE.
+ Check out the [command-line tutorial][godot-command-line] for more information.
+```
```
Bromeon: This shouldn't be a warning, but rather a tip.

Warnings are to prevent errors -- this is more of an input to become a more efficient developer.

Bromeon (commented):
Thanks a lot!

Instead of adding a note to the now-outdated text, could you please adjust the text so that it reflects the status quo? The workaround should no longer be listed. You can still mention that by default, GDExtension runs all Rust classes, but that gdext explicitly changes this behavior.

Also, instead of linking to the PR, you could link to the `ExtensionLibrary` API docs:
https://godot-rust.github.io/docs/gdext/master/godot/init/trait.ExtensionLibrary.html

Please also use footnote style for links: `[Some text][id]` and on the bottom `[id]: link` (see existing files).

RefinedDev:
Apologies for not using footnote-style for links, I have edited the 'Always-on' paragraph as per the status quo and changed the warning into a tip.

src/gdext/index.md
```diff
@@ -36,19 +36,7 @@ That said, there are some notable differences:
...
-   if Engine::singleton().is_editor_hint() {
-      return;
-   }
-   ```
-
-   ```admonish note
-   This is **no longer necessary**. By default, all virtual callbacks (ready, process etc.) are no longer invoked.
-   [More information here.](https://github.com/godot-rust/gdext/pull/365)
-   ```
+   There is no differentiation between "tool" and "normal" scripts anymore; Rust logic runs as soon as the Godot editor launches, but gdext explicitly changes this behaviour, by default all virtual callbacks (ready, process etc.) are not invoked **in editor mode**, this behaviour can be reverted with a configuration, while implementing the [ExtensionLibrary][extension_library_doc] trait.
```
Bromeon: Maybe use some full stops, otherwise this sentence is very long. You can then split some lines, too.
Also, we use American english ("behavior" instead of "behaviour"). Any code symbols I would put inside backticks. 🙂 

```suggestion
   There is no differentiation between "tool" and "normal" scripts anymore, as it was the case in GDNative. Rust logic runs as soon as 
   the Godot editor launches, but gdext explicitly changes this behavior. By default, all virtual callbacks (`ready`, `process` etc.)
   are not invoked **in editor mode**. This behavior can be configured when implementing the [`ExtensionLibrary`][extension-library-doc] trait.
```

src/gdext/index.md
```diff
@@ -58,3 +46,4 @@ That said, there are some notable differences:
 
 [features]: https://github.com/godot-rust/gdextension/issues/24
 [issue #66231]: https://github.com/godotengine/godot/issues/66231
+[extension_library_doc]: https://godot-rust.github.io/docs/gdext/master/godot/init/trait.ExtensionLibrary.html#method.editor_run_behavior
```
Bromeon: Minor nitpick for consistency:
```suggestion
[extension-library-doc]: https://godot-rust.github.io/docs/gdext/master/godot/init/trait.ExtensionLibrary.html#method.editor_run_behavior
```

Bromeon (commented):
Thanks for the update, no need to apologize! 🙂
A few more comments, then should be good to go!

RefinedDev:
Thank you, and sorry for the inconvenience about the grammar. Are we good to go now?

Bromeon:
Thank you!

PR #110, opened 2023-08-05
Updated the path to the the compiled dynamic library on macOS in hell…

RamziA961:
Currently, hello-world.md's gdextension example does not include the prefix 'lib' before the crate name for the compiled dynamic library paths on macOS. I updated the file to change the paths for 'macos.debug', 'macos.release', 'macos.debug.arm64', 'macos.release.arm64' to include it.


Bromeon:
Nice catch, thanks for the PR!

I checked and indeed, [we use the `lib` prefix in `itest`](https://github.com/godot-rust/gdext/blob/master/itest/godot/itest.gdextension), which is run in CI 🙂

PR #111, opened 2023-08-12
Inline code fix for hello-world.md

ogapo:
Recent change in master makes emit_signal only available on the base.

Caught this while following along with the tutorial, figured I'd help out where I can. Great work btw, the rust API feels pretty ergonomic so far while still accommodating the non-Rusty patterns of a C++ game engine.


Bromeon:
Thanks a lot!

PR #112, opened 2023-08-12
Create debugging.md

ogapo:
added a basic summary of debug setup assuming VSCode IDE

### DIFF
--- a/src/gdext/advanced/debugging.md
+++ b/src/gdext/advanced/debugging.md
@@ -1,0 +1,75 @@
+# Debugging
+
+You can debug your Rust code simply by attaching LLDB to the Godot editor and telling the editor to launch your project. 
+However, the process for doing this will vary based on your IDE and platform.
+
+## Launching with VSCode
+
+Here is an example launch configuration for Visual Studio Code (typically found in `./.vscode/launch.json` relative to your project's root). This example assumes you have the [CodeLLDB](https://marketplace.visualstudio.com/items?itemName=vadimcn.vscode-lldb) 
+extension installed which is common for Rust development.
+
+```jsonc
+{
+    "configurations": [
+        {
+            "name": "Debug Project (Godot 4)",
+            "type": "lldb",
+            "request": "launch",
+            "cwd": "${workspaceFolder}",
+            "args": [
+                "-e", // run editor (remove this if you want to launch the scene directly)
+                "-w", // windowed mode
+            ],
+            "linux": {
+                "program": "/usr/local/bin/godot4",
+            },
+            "windows": {
+                "program": "C:\\Program Files\\Godot\\Godot_v4.1.X.exe",
+            },
+            "osx": {
+                // NOTE: on Mac the Godot.app needs to be manually re-signed to enable debugging (see below)
+                "program": "/Applications/Godot.app/Contents/MacOS/Godot",
+            }
+        }
+    ]
+}
+```
+
+## Debugging on MacOS
+
+Attaching a debugger to an executable that wasn't compiled locally (the Godot editor, in this example) requires special considerations on MacOS due to its System 
+Integrity Protection (SIP) security feature. 
+Even though your extension will be compiled locally, LLDB will be unable to attach to the Godot host process without manual resigning.
+
+In order to re-sign, simply create a file called `editor.entitlements`[^1] with the following contents (it is recommended to check this file in since each dev will need to resign their local install if you have a team):
+
+[^1]: Be sure to use the `editor.entitlements` file below rather than the one from the Godot docs as it includes the required `com.apple.security.get-task-allow` key not currently present in Godot's instructions.
+
+```xml
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+    <dict>
+        <key>com.apple.security.cs.allow-dyld-environment-variables</key>
+        <true/>
+        <key>com.apple.security.cs.allow-jit</key>
+        <true/>
+        <key>com.apple.security.cs.allow-unsigned-executable-memory</key>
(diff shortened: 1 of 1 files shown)


src/gdext/advanced/debugging.md
```diff
@@ -0,0 +1,75 @@
+# Debugging
+
+You can debug your Rust code simply by attaching LLDB to the Godot editor and telling the editor to launch your project. 
+However, the process for doing this will vary based on your IDE and platform.
```
Bromeon: Is LLDB not already platform-specific?

Also, using the editor is not a requirement. One can start the Godot engine (without editor) directly, depending on the setup.

src/gdext/advanced/debugging.md
```diff
@@ -0,0 +1,75 @@
...
+}
+```
+
+## Debugging on MacOS
+
+Attaching a debugger to an executable that wasn't compiled locally (the Godot editor, in this example) requires special considerations on MacOS due to its System 
+Integrity Protection (SIP) security feature. 
+Even though your extension will be compiled locally, LLDB will be unable to attach to the Godot host process without manual resigning.
+
+In order to re-sign, simply create a file called `editor.entitlements`[^1] with the following contents (it is recommended to check this file in since each dev will need to resign their local install if you have a team):
```
Bromeon: "resign" vs. "re-sign" -- I'd consistently use the latter, as the former is a verb on its own with different meaning.

Also, maybe avoid abbreviations like "dev", and put a comma before "since each" 🙂

src/gdext/advanced/debugging.md
```diff
@@ -0,0 +1,75 @@
+# Debugging
+
+You can debug your Rust code simply by attaching LLDB to the Godot editor and telling the editor to launch your project. 
+However, the process for doing this will vary based on your IDE and platform.
+
+## Launching with VSCode
+
+Here is an example launch configuration for Visual Studio Code (typically found in `./.vscode/launch.json` relative to your project's root). This example assumes you have the [CodeLLDB](https://marketplace.visualstudio.com/items?itemName=vadimcn.vscode-lldb) 
+extension installed which is common for Rust development.
```
Bromeon:
```suggestion
extension installed, which is common for Rust development.
```

src/gdext/advanced/debugging.md
```diff
@@ -0,0 +1,75 @@
...
+        <true/>
+        <key>com.apple.security.get-task-allow</key>
+        <true/>
+    </dict>
+</plist>
+```
+
+Then you can run `codesign -s - --deep --force --options=runtime --entitlements ./editor.entitlements /Applications/Godot.app` in Terminal to complete the process.
+
+The resigning process is covered in more detail in the [Godot Docs](https://docs.godotengine.org/en/stable/contributing/development/debugging/macos_debug.html).
```
Bromeon:
```suggestion
The re-signing process is covered in more detail in the [Godot Docs](https://docs.godotengine.org/en/stable/contributing/development/debugging/macos_debug.html).
```

Bromeon (commented):
Thanks a lot, that's a great addition 👍

src/gdext/advanced/debugging.md
ogapo: I guess I just meant the Mac instructions are different for windows, even though LLDB is on all the platforms.

I believe if you don't pass `-e` in the launch.json file config then it won't boot into the editor. Is that what you mean? I can call that out explicitly if it's helpful.

Bromeon: I meant that 
> You can debug your Rust code simply by attaching LLDB to the Godot editor

sounds like you need to attach the debugger to the Godot editor, which is not the case. It's also possible to launch the application directly from the IDE (although I'm not familiar with VS Code here).

So I thought maybe mention that the editor is not explicitly _required_.
I saw that you added
```jsonc
                "-e", // run editor (remove this if you want to launch the scene directly)
```
which is nice and should be enough! 🙂

src/gdext/advanced/debugging.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Debugging on MacOS
+
+Attaching a debugger to an executable that wasn't compiled locally (the Godot editor, in this example) requires special considerations on MacOS due to its System 
+Integrity Protection (SIP) security feature. 
+Even though your extension will be compiled locally, LLDB will be unable to attach to the Godot host process without manual re-signing.
+
+In order to re-sign, simply create a file called `editor.entitlements`[^1] with the following contents (it is recommended to check this file in since each dev will need to re-sign their local install if you have a team):
+
+[^1]: Be sure to use the `editor.entitlements` file below rather than the one from the Godot docs as it includes the required `com.apple.security.get-task-allow` key not currently present in Godot's instructions.
```
Bromeon: Should this really be a foot note? In that case, "below" is no longer accurate.
Maybe you can add an "admonish note" (see other pages in the book) instead.

Also, comma before "as it includes" 😉

ogapo: good point, will put it into the text itself

Bromeon:
Thanks a lot!

Edit: I noticed we forgot the link in the summary page -- I added it on `master`. Also fixed a few more things like "macOS", formatting etc.

ogapo:
@Bromeon I forgot to add 
```
            "preLaunchTask": "rust: cargo build",
```
to the example so it will ensure the rust code is built up to date before launching the debugger. It works as written, but I think that's probably the behavior most people would likely expect. Not sure it's worth a PR if you wanna make that update or not.

Bromeon:
Added it; is it good the way it is deployed now? 
https://godot-rust.github.io/book/gdext/advanced/debugging.html

ogapo:
I've been messing with my own some today. I ended up adding `"sourceLanguages": ["rust"],` and commenting out both the -e and -w argument. "-e" seems to prevent LLDB from being able to trigger some breakpoints, but I'm not sure why. The same breakpoints hit ok without the -e though. I think it may be something to do with how we are trapping Rust panics rather than letting them signal but I couldn't figure out how.

PR #114, opened 2024-02-29
Updating shell.nix to flakes

jaoleal:
Updating the shell.nix to a flake.nix. 
See the Issue on gdnative for more understanding. 

https://github.com/godot-rust/gdnative/pull/1069#issuecomment-1971001997

### DIFF
--- a/src/recipes/nix-build-system.md
+++ b/src/recipes/nix-build-system.md
@@ -66,4 +66,4 @@ If you get any errors about missing headers, you can use [`nix-locate`](https://
 
 ## Activating the Flake environment
 
-In the same directory that you made your flake.nix, you can use the comand ['nix develop'](https://nixos.wiki/wiki/Development_environment_with_nix-shell#nix-develop) to load our flake config and setup everything for your NixOs system.
+In the same directory where you created your flake.nix, you can use the comamnd [`nix develop`](https://nixos.wiki/wiki/Development_environment_with_nix-shell#nix-develop) to load our Flake config and setup everything for your NixOS system.
--- a/src/recipes/nix-build-system.md
+++ b/src/recipes/nix-build-system.md
@@ -1,4 +1,4 @@
-# Recipe: Flakes as development environment
+# Recipe: Nix Flakes as development environment
 
 **Disclaimer**: _Currently the following steps are tested and confirmed to work on NixOS only._
 
--- a/src/recipes/nix-build-system.md
+++ b/src/recipes/nix-build-system.md
@@ -2,13 +2,28 @@
 
 **Disclaimer**: _Currently the following steps are tested and confirmed to work on NixOS only._
 
-[Flakes](https://nixos.wiki/wiki/Flakes) Flakes is a feature of managing Nix packages to simplify usability and improve reproducibility of Nix installations.
+[Flakes](https://nixos.wiki/wiki/Flakes) Flakes is a widelly used experimental feature of managing Nix packages to simplify usability and improve reproducibility of Nix installations.
 
-This tutorial assumes that you are on a [NixOs](https://nixos.wiki/wiki/Main_Page) system
-or that Nix is [installed](https://nixos.org/download.html#nix-quick-install).
-Finally, you also need to make sure that [Flakes](https://nixos.wiki/wiki/Flakes) is enabled.
+Please read carefully all this page before starting to use gdnative on NixOS.
 
-To begin with, we are going to create a new project using the [Hello, world!](../getting-started/hello-world.md) guide in Getting Started. Note that the full source code for the project is available at https://github.com/godot-rust/godot-rust/tree/master/examples/hello-world. Because we aren't using the default build system explained in [setup](../getting-started/setup.md), you should only be worried about the content of the project rather than the dependencies.
+This tutorial assumes that you are on a [NixOs](https://nixos.wiki/wiki/Main_Page) system or that Nix is [installed](https://nixos.org/download.html#nix-quick-install).
+And so on, you also need to make sure that [Flakes](https://nixos.wiki/wiki/Flakes) is enabled.
+
+To begin with, follow the [Hello, world!](../getting-started/hello-world.md) guide in Getting Started. Note that the full source code for the project is available at https://github.com/godot-rust/godot-rust/tree/master/examples/hello-world. Because we aren't using the default build system explained in [setup](../getting-started/setup.md), you should only be worried about the content of the project rather than the dependencies.
+
+## What we need ?
+
+  Since [godot-rust](https://github.com/godot-rust/gdnative) use [gdnative](https://docs.godotengine.org/en/3.5/tutorials/scripting/gdnative/what_is_gdnative.html) to bind Rust and Godot, we need to install the following dependencies:
+
+  - The [rust-overlay](https://github.com/oxalica/rust-overlay) to include some rusty toolchains like `rustc`, `cargo`, `rustup` and `rls`.
+  - [openssl](https://www.openssl.org/) that provide us some cryptographic functions for rust.
+  - [pkg-config](https://www.freedesktop.org/wiki/Software/pkg-config/) to manage the compilation of the project.
+  - [godot3](https://godotengine.org/) aliased as godot.
+  - [clang](https://clang.llvm.org/) to compile and bind rust code to gdnative.
+
+  We will use the unstable version of nixpkgs to get the latest version of all packages and the flakes feature.
+
+  And we need to specify the `LIBCLANG_PATH` and `BINDGEN_EXTRA_CLANG_ARGS` to make the gdnative aware of the headers and the library.
 
 
 ## Specifying dependencies
@@ -66,9 +81,9 @@ Now to the Nix part of the tutorial. In your project directory create a `flake.n
 }
 ```
 
-If you get any errors about missing headers, you can use [`nix-locate`](https://github.com/bennofs/nix-index#usage) to search for them, e.g. `nix-locate 'include/wchar.h' | grep -v '^('` (the `grep -v` hides indirect packages), and then add the matching Nix package via the `BINDGEN_EXTRA_CLANG_ARGS` env var like above ([context](https://github.com/NixOS/nixpkgs/issues/52447#issuecomment-853429315)).
 
 
 ## Activating the Flakes environment
 
 In the same directory where you created your flake.nix, you can use the command [`nix develop`](https://nixos.wiki/wiki/Development_environment_with_nix-shell#nix-develop) to load our Flakes config and setup everything for your Nix-capable system.
+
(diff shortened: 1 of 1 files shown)


src/recipes/nix-build-system.md
```diff
@@ -1,71 +1,69 @@
 # Recipe: Nix as development environment
 
-**Disclaimer**: _Currently the following steps are tested and confirmed to work on Linux only._
+**Disclaimer**: _Currently the following steps are tested and confirmed to work on NixOs only._
 
-[Nix](https://nixos.org/) is a package manager that employs a pure functional approach to dependency management. Nix packages are built and ran in isolated environments. It makes them more portable, but also harder to author. This tutorial will walk you through the process of setting up a package for Godot, GDNative and Rust with Nix.
```
Bromeon, line 5: Why did you remove the introduction about NixOS? I don't think Flakes should be explained before NixOS itself.

src/recipes/nix-build-system.md
```diff
@@ -1,71 +1,69 @@
 # Recipe: Nix as development environment
 
-**Disclaimer**: _Currently the following steps are tested and confirmed to work on Linux only._
+**Disclaimer**: _Currently the following steps are tested and confirmed to work on NixOs only._
```
Bromeon:
```suggestion
**Disclaimer**: _Currently the following steps are tested and confirmed to work on NixOS only._
```

Bromeon: Capitalization is also incorrect in some other places in this page, maybe search & replace 🙂

src/recipes/nix-build-system.md
```diff
@@ -1,71 +1,69 @@
...
+## Activating the Flake environment
 
-To run the project, first you need to compile the `hello-world` Rust library using `cargo build`. After that, you can open the Godot Engine in your terminal using the command `godot`. As seen in `shell.nix`, this command is actually aliased to `nixGL godot -e` in which Godot will be opened using nixGL instead of opening it directly. After running the default scene, you should be able to see a single `hello, world.` printed in the Godot terminal.
+In the same directory that you made your flake.nix, you can use the comand ['nix develop'](https://nixos.wiki/wiki/Development_environment_with_nix-shell#nix-develop) to load our flake config and setup everything for your NixOs system.
```
Bromeon:
```suggestion
In the same directory where you created your flake.nix, you can use the comamnd [`nix develop`](https://nixos.wiki/wiki/Development_environment_with_nix-shell#nix-develop) to load our Flake config and setup everything for your NixOS system.
```

Bromeon (commented):
Thanks!

I'm not using NixOS myself, so I'll need some other reviewers to check on the content and rationale.

src/recipes/nix-build-system.md
jaoleal: a little redundant since the relation is with flakes feature not nix package manager. But this doesnt change nothing after all

jaoleal:
Actually, im not a experienced user of NixOs, flakes and all that good stuff. I prefer that some other folks check the content too

src/recipes/nix-build-system.md
```diff
@@ -1,71 +1,69 @@
...
+## Activating the Flakes environment
 
-To run the project, first you need to compile the `hello-world` Rust library using `cargo build`. After that, you can open the Godot Engine in your terminal using the command `godot`. As seen in `shell.nix`, this command is actually aliased to `nixGL godot -e` in which Godot will be opened using nixGL instead of opening it directly. After running the default scene, you should be able to see a single `hello, world.` printed in the Godot terminal.
+In the same directory where you created your flake.nix, you can use the comamnd [`nix develop`](https://nixos.wiki/wiki/Development_environment_with_nix-shell#nix-develop) to load our Flakes config and setup everything for your NixOS system.
```
storopoli:
```suggestion
In the same directory where you created your flake.nix, you can use the command [`nix develop`](https://nixos.wiki/wiki/Development_environment_with_nix-shell#nix-develop) to load our Flakes config and setup everything for your Nix-capable system.
```

Nix is available in other distros as well

src/recipes/nix-build-system.md
```diff
@@ -1,71 +1,69 @@
...
+# Recipe: Flakes as development environment
 
-**Disclaimer**: _Currently the following steps are tested and confirmed to work on Linux only._
+**Disclaimer**: _Currently the following steps are tested and confirmed to work on NixOS only._
 
-[Nix](https://nixos.org/) is a package manager that employs a pure functional approach to dependency management. Nix packages are built and ran in isolated environments. It makes them more portable, but also harder to author. This tutorial will walk you through the process of setting up a package for Godot, GDNative and Rust with Nix.
+[Flakes](https://nixos.wiki/wiki/Flakes) Flakes is a feature of managing Nix packages to simplify usability and improve reproducibility of Nix installations.
 
-This tutorial assumes that Nix is [installed](https://nixos.org/download.html#nix-quick-install) in your system.
+This tutorial assumes that you are on a [NixOs](https://nixos.wiki/wiki/Main_Page) system and have [Flakes](https://nixos.wiki/wiki/Flakes) enabled.
```
storopoli:
```suggestion
This tutorial assumes that you are on a [NixOs](https://nixos.wiki/wiki/Main_Page) system
or that Nix is [installed](https://nixos.org/download.html#nix-quick-install).
Finally, you also need to make sure that [Flakes](https://nixos.wiki/wiki/Flakes) is enabled.
```

src/recipes/nix-build-system.md
storopoli: We shouldn't remove this.

src/recipes/nix-build-system.md
```diff
@@ -1,71 +1,69 @@
-# Recipe: Nix as development environment
+# Recipe: Flakes as development environment
```
storopoli:
```suggestion
# Recipe: Nix Flakes as development environment
```

src/recipes/nix-build-system.md
jaoleal: Yes, but i didnt include that because of nixgl.
Nixgl is some graphics drivers workarounds for non-nixos nix package manager and it looks like they does not have a good flakes support.

src/recipes/nix-build-system.md
jaoleal: > Yes, but i didnt include that because of nixgl. Nixgl is some graphics drivers workarounds for non-nixos nix package manager and it looks like they does not have a good flakes support.

This again.

jaoleal:
Im commting the nix on non-nixos changes. May the Nixgl is not necessary at all.

chitoyuu:
Should the old content really be replaced altogether? On https://nixos.wiki/wiki/Flakes it is stated that (emphasis mine):

> Nix flakes is an *[experimental feature](https://nixos.org/manual/nix/stable/contributing/experimental-features.html)* of the Nix package manager.

Not everyone might want to enable experimental features for their package manager, and not everyone is using NixOS (instead of non-NixOS Nix environments, which the previous author apparently was on, and this setup is stated to not support). I'm not sure if the old directions should be removed, if it isn't superseded entirely.

PgBiel:
Nix flakes are pretty widely used nowadays despite that warning. Even then, I agree we should keep previous instructions regarding `default.nix` and `shell.nix` (& co.) if they were already there. I'd be more in favor of having a separate section, e.g. "Using Flakes", with an alternative setup with flakes. They do deserve a mention, but that doesn't mean we have to remove or deprecate prior instructions.
(Disclaimer: I am a NixOS user, but I'm not a full expert on the subject.)

Of note, unless we are doing something with some specific cursed dependencies, it's very likely a Nix setup - with flakes or otherwise - would work on non-NixOS environments as well. At least, it's worth testing in other distros with Nix installed (e.g. Ubuntu, Fedora, OpenSUSE).

**EDIT:** From a quick look at the PR, it seems that NixGL is used, so I _think_ there might be problems using this particular setup outside of NixOS, but I'm not sure. Worth testing.

storopoli:
> Even then, I agree we should keep previous instructions regarding `default.nix` and `shell.nix` (& co.) if they were already there. I'd be more in favor of having a separate section, e.g. "Using Flakes", with an alternative setup with flakes.

If you have a Flake.nix it is trivial to extend that to a Nix shell.
The problem with Nix shells (outside of a Flake, i.e. `default.nix` or `shell.nix`) is that they are not reproducible and are tied to the users config and nixpkgs.

PgBiel:
> If you have a Flake.nix it is trivial to extend that to a Nix shell.

True, you have a point. You can use `flake-compat` to convert a Flake to the older format.
Then, for the sake of maintainability, it might be better to just present the flake setup, and suggest using `flake-compat` if flakes are not enabled.

Alternatively, we can have some more high-level instructions indicating which packages you need to use `gdnative` with Nix, and then use the flake as an example.

storopoli:
@jaoleal see if you understand these suggestions.
I think that they are a good way forward for this PR.
I can help you clarify anything if you need.

jaoleal:
> Should the old content really be replaced altogether?

Well, thinking that is good if we have options to use on Nix... I think that we have to maintain those that work and not hold on to outdated technologies or ways to do the same thing. In every context that i used that file, i had problems.

jaoleal:
> Of note, unless we are doing something with some specific cursed dependencies, it's very likely a Nix setup - with flakes or otherwise - would work on non-NixOS environments as well. At least, it's worth testing in other distros with Nix installed (e.g. Ubuntu, Fedora, OpenSUSE).

So, the non-NixOs environment... In that context, as the previous docs say, we could have some problems with opengl and then we could use NixGl that... is something I couldn't get to work on my environment until now, i`ll give another try.
If my speak sound confusing, feel free to ask me again.

PgBiel:
> > Of note, unless we are doing something with some specific cursed dependencies, it's very likely a Nix setup - with flakes or otherwise - would work on non-NixOS environments as well. At least, it's worth testing in other distros with Nix installed (e.g. Ubuntu, Fedora, OpenSUSE).
> 
> So, the non-NixOs environment... In that context, as the previous docs say, we could have some problems with opengl and then we could use NixGl that... is something I couldn't get to work on my environment until now, i`ll give another try. If my speak sound confusing, feel free to ask me again.

No you're right, I hadn't seen the NixGL dependency. That one can definitely be a pain outside of NixOS I believe. So it's fair enough.

I do, however, wonder if there's some escape hatch which would allow us to not depend on NixGL for non-NixOS users. But that will suffice for the goal of this PR.

jaoleal:
> I do, however, wonder if there's some escape hatch which would allow us to not depend on NixGL for non-NixOS users. But that will suffice for the goal of this PR.

The goal of this PR is that anyone have to spend some time on managing painfull deps. ill be happy to ensure that non-NixOs nix users have they time saved but i cannot absolutely sure that it will work... i only have a NixOs machine. 
Anyway, ill try to include NixGl on the party. :)

PR #115, opened 2024-04-23
Spelling Correction in traits.md

NoodleLadJr:
Fixed 'wouuld' to 'would'


Bromeon:
Thanks! Could you also add single backticks around `cargo check` (so it shows in monospace)?

NoodleLadJr:
Believe I've taken care of it now

Bromeon:
Thanks a lot!

PR #116, opened 2025-06-23
Fixing a variety of broken links

reshen:
Just a small url fix here.

### DIFF
--- a/src/bind/classes.md
+++ b/src/bind/classes.md
@@ -33,21 +33,21 @@ fn init(handle: InitHandle) {
 
 ## Class definition
 
-Similar to the [Hello World](../getting-started/hello-world.md#overriding-a-godot-method) example, we can define the `GodotApi` native class as follows:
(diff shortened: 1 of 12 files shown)


Bromeon:
Thanks!

The fix is incorrect, gdnative targets Godot 3, while "stable" documents Godot 4.
It should be:
https://docs.godotengine.org/en/3.6/tutorials/performance/threads/thread_safe_apis.html

Can you see if other such links are affected?

reshen:
I used [Michael-F-Bryan/mdbook-linkcheck](https://github.com/Michael-F-Bryan/mdbook-linkcheck) to find the majority of these broken links, but decided against automating the check because (a) it's full of false positives, (b) I couldn't find a way to reduce the errors to merely warnings; the former breaks our build which doesn't work for us.  At any rate, I believe these links are sound now, lmk if you spot any other issues in the PR. :)

Bromeon:
Thank you so much! :heart:

